PR #34, opened 2024-02-06
Bump `lean-toolchain`

0art0:


david-christiansen:
I think that #45 will end up superceding this PR, as it adds a dependency on a nightly for a while (to get an essential feature).

Thanks, though!

PR #78, opened 2024-04-18
feat: incremental elaboration of Verso documents

david-christiansen:
Based on https://github.com/leanprover/lean4/pull/3849

The API is still a bit fiddly, so this code needs some cleanup prior to merge


david-christiansen:
@Kha - I think this is still using the framework in the intended manner, though I lifted out the hairy parts to a helper. I'm not totally satisfied by how the different monads interact, but couldn't immediately think of a nicer way.

Is this about how you'd imagined it being, or am I missing some low-hanging fruit for simplification?

src/verso/Verso/Doc/Elab/Incremental.lean
```diff
@@ -0,0 +1,209 @@
+/-
+Copyright (c) 2023-2024 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: David Thrane Christiansen
+-/
+
+import Lean
```
Kha: Temporary coarse import?

src/verso/Verso/Doc/Elab/Incremental.lean
```diff
@@ -0,0 +1,209 @@
...
+  if let some snap := (← read).snap? then
+    withReader (fun ρ => {ρ with snap? := none}) do
+      lift <| do
+        let mut oldSnap? := snap.old?.bind (·.val.get.toTyped? (α := snapshot))
+        let mut nextPromise ← IO.Promise.new
+        let initData := {diagnostics := .empty}
+        snap.new.resolve <| DynamicSnapshot.ofTyped <|
+          mkSnap cmd initData (.pure (← get)) (some {range? := cmd.getRange?, task := nextPromise.result})
+        initAct
+        let mut errors : MessageLog := .empty
```
Kha: I think we should be able to use `m`'s message log instead of keeping track of it manually. I'm not even sure the `finally` does the right thing for local mutable variables.

src/verso/Verso/Doc/Concrete.lean
```diff
@@ -168,6 +171,61 @@ macro "%docName" moduleName:ident : term =>
...
+deriving Nonempty
+
+structure DocElabSnapshotData extends Language.Snapshot where
+  stx      : Syntax
+  finished : Task DocElabSnapshotState
+deriving Nonempty
+
+inductive DocElabSnapshot where
+  | mk (data : DocElabSnapshotData) (next : Option (SnapshotTask DocElabSnapshot))
+deriving TypeName, Nonempty
```
Kha: I think we could bake this part into the `incrementallyElabCommand` API in order to reduce the number of required accessor parameters without really losing generality. Or rather, we could if it wasn't for `TypeName`?

src/verso/Verso/Doc/Elab/Incremental.lean
```diff
@@ -0,0 +1,209 @@
...
+    (lift : {α : _} → m α → CommandElabM α)
+    : CommandElabM (IO.Promise snapshot × σ) := do
+  if let some snap := (← read).snap? then
+    withReader (fun ρ => {ρ with snap? := none}) do
+      lift <| do
+        let mut oldSnap? := snap.old?.bind (·.val.get.toTyped? (α := snapshot))
+        let mut nextPromise ← IO.Promise.new
+        let initData := {diagnostics := .empty}
+        snap.new.resolve <| DynamicSnapshot.ofTyped <|
+          mkSnap cmd initData (.pure (← get)) (some {range? := cmd.getRange?, task := nextPromise.result})
```
Kha: `cmd` can probably be replaced with `mkNullNode steps` since we really only care about the union of ranges

david-christiansen:
Note to self after call: check `getAndResetMessageLog`, that avoids double-reported diagnostics from incrementality

src/verso/Verso/Doc/Elab/Incremental.lean
david-christiansen: OK - this has been fixed in the latest version. Thanks!

src/verso/Verso/Doc/Elab/Incremental.lean
david-christiansen: OK. Done :)

src/verso/Verso/Doc/Concrete.lean
david-christiansen: I got about 70% of the way there.

david-christiansen:
OK, I think this is as ready as it's getting, though some items may not work until https://github.com/leanprover/lean4/pull/3940. Once that's in, I'll try again, but this is already better than it was.

PR #125, opened 2024-07-17
fix: leading zero for single-digit days and months in ISO 8601 dates

TwoFX:
I hope this PR is welcome.

To the best of my knowledge, 2024-7-17 is not a valid ISO 8601 date (I don't have access to ISO 8601 itself, the closest I could come was RFC 3339, according to which it is not a valid date). Of course I don't claim that my proposed change is anything close to a proper date formatting routine, but it should ensure that at least all of the dates on the Lean blog are valid. With Time.lean moving to core, we will hopefully have access to a good formatter in the not-too-distant future. The only reason why I'm proposing to change this beforehand is because it keeps annoying me whenever I open the Lean blog :)

Post URLs are unaffected by this change.

### DIFF
--- a/src/verso-blog/Verso/Genre/Blog/Basic.lean
+++ b/src/verso-blog/Verso/Genre/Blog/Basic.lean
@@ -97,6 +97,14 @@ structure Date where
   day : Nat
 deriving Inhabited, Repr
 
+-- FIXME replace this primitive implementation with proper ISO 8601
+-- date formatting when available
+def Date.toIso8601String (date : Date) : String :=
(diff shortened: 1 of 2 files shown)


david-christiansen:
It wasn't intended to be ISO dates - it's just the way I happen to usually write down dates in a cross-cultural context 😄 

But if this annoys you, it likely annoys others, so the default should change!

david-christiansen:
I also wish static site generators had a nice way to set up redirects, because it would be good to be able to fix up URLs later.

nomeata:
If this is on netlify you can generate `_redirects` file: https://docs.netlify.com/routing/redirects/

PR #143, opened 2024-08-12
fix: incremental message logs - off-by-one fix and footgun removal

david-christiansen:
Fixes incremental reporting, which previously suffered from two problems:

1. Snapshot message logs were restored only when states were to be re-used, but message logs capture the log at the _start_ of the elaboration step and thus should be unconditionally restored. This led to the step immediately prior to buffer damage losing its messages.

2. It was inconvenient to disable incrementality for "finishing" commands, so it didn't happen in practice. Now, the finishing command is provided as a parameter to incrementallyElabCommand, so the framework can arrange for it to be run correctly and the details of promises that need resolving are not exposed. This led to the final messages getting lost, while now they're saved. Footgun removal!

### DIFF
--- a/src/verso/Verso/Doc/Concrete.lean
+++ b/src/verso/Verso/Doc/Concrete.lean
@@ -205,22 +205,19 @@ elab (name := completeDoc) "#doc" "(" genre:term ")" title:inlineStr "=>" text:c
       | dbg_trace "nope {ppSyntax title}" throwUnsupportedSyntax
     let titleString := inlinesToString (← getEnv) titleParts
     let initState : PartElabM.State := .init titleName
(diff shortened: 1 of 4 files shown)


david-christiansen:
This is still a draft for two reasons:
 1. It would be even better to save the post-step message log in the snapshot
 2. The snapshot also needs to capture more state from the underlying monad that is probably getting lost right now

david-christiansen:
This code seems to work correctly, because the ways that it's not yet right aren't exercised by Verso. Fixing it properly will require more time than I have right now, so I took notes and saved the branch. Merging this will fix UX issues that have been causing more and more difficulties for me.

PR #148, opened 2024-08-20
fix: work around go-to-definition limitation in language server

david-christiansen:
The language server presently assumes that each module path has a unique source directory corresponding to it. This rearranges Verso's internals so that this is the case.

Imports of Verso.Genre.Manual should now be VersoManual, and Verso.Genre.Blog should now be VersoBlog.


david-christiansen:
This is a workaround for https://github.com/leanprover/lean4/issues/4962

Issue #174, opened 2024-09-23
Manual genre: adopt Scribble-style nav bar on left


david-christiansen:
Scribble has a much nicer local ToC view than VersoManual does right now. We should adopt something similar.

Screenshot:
![image](https://github.com/user-attachments/assets/f05fcc7b-890e-46ce-8c38-510ff109e3d4)

Advantages include:
 * Nesting doesn't lead to width
 * There's a place to put index entries as well as section headers

david-christiansen:
Fixed in #186

Issue #173, opened 2024-09-23
Manual genre: more consistent section numbering


david-christiansen:
Right now, CSS is used to number sections. But this makes it more difficult to show the numbers properly in the section headers, which would make the document easier to navigate.

It would be better to emit the section numbers in the HTML directly, which would also make numbered section references possible (e.g. "see section 1.3.5").

david-christiansen:
Implemented in #186

PR #186, opened 2024-10-18
Manual genre: generate section numbers in HTML and use a Scribble-style left margin ToC

david-christiansen:
This makes it much easier to navigate complicated documents, and provides an automatic breadcrumb feature.


david-christiansen:
Fixes #174 and #173

PR #210, opened 2024-10-30
fix: adhere much more closely to the HTML spec for attribute names

david-christiansen:
This allows attribute names like `aria-busy` and `data-foo` to be included directly in HTML templates.


jakobvase:
Nice :)

PR #211, opened 2024-10-30
fix: double scrolling in chrome and horizontal scrolling

frangio:
Fixes https://github.com/leanprover/reference-manual/issues/128 by adding `position: relative`.

Unfortunately not very clear why this works.

---

This PR also adds some horizontal padding to prevent horizontal scrolling from widgets in the margins like permalinks. (Although there's also horizontal scrolling due to long `code` lines that should be fixed separately.)

image

### DIFF
--- a/src/verso-manual/VersoManual/Html/Style.lean
+++ b/src/verso-manual/VersoManual/Html/Style.lean
@@ -442,6 +442,7 @@ main .authors {
 
 main > section {
     margin: auto;
+    position: relative;
 }
 
 main section {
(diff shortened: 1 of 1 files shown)


david-christiansen:
The ways of CSS are mysterious :-) Thanks!

As far as the padding goes, the variable is in the area of the file that's intended for eventual theme customization. Is this something that we'd expect a theme (e.g. dark mode) to override? It seems not, to me, but I'd like to hear your thoughts before just moving it elsewhere.

frangio:
Ah, I agree it's not really something to be themed. I just wanted to use a variable to make sure the padding was enough for the permalink, but feel free to inline the variable or move it somewhere!

david-christiansen:
#218 contains these changes plus the dicussed update. Thanks again!

PR #224, opened 2024-11-10
fix: use better mobile CSS for highlighted code and in manuals

david-christiansen:
* Switch to rems universally
 * Include the magic "I know what I'm doing" meta tag

Based on advice in https://github.com/leanprover/reference-manual/issues/140 and https://github.com/leanprover/reference-manual/issues/137. It won't solve all the issues, and it requires a bit more testing, but a quick check was positive.

### DIFF
--- a/src/verso-manual/VersoManual/Html/Style.lean
+++ b/src/verso-manual/VersoManual/Html/Style.lean
@@ -92,17 +92,12 @@ pre, code {
 /******** Page Layout ********/
 
 .with-toc #toc {
-    height: 100vh;
     position: fixed;
     z-index: 10;
 }
 
 /** Non-mobile **/
 @media screen and (min-width: 701px) {
-    .with-toc #toc {
-        height: 100vh;
-    }
-
     .with-toc > main {
         /* NB main > section also has padding that's added to this in practice */
         padding-left: var(--verso-toc-width);
@@ -128,7 +123,7 @@ pre, code {
     display: flex;
     flex-direction: column;
     justify-content: space-between;
-    height: 100vh;
+    height: 100dvh;
     width: var(--verso-toc-width);
 }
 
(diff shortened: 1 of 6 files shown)


src/verso-manual/VersoManual/Html/Style.lean
```diff
@@ -74,46 +86,35 @@ dd > p:first-child {
...
-.with-toc #top-menu {
-    grid-area: toggle;
+    position: fixed;
+    z-index: 10;
 }
 
-.with-toc #toc {
-    overflow-y: auto;
+/** Non-mobile **/
+@media screen and (min-width: 701px) {
```
jakobvase, line 100: I would not use both min-width and max-width across a stylesheet. In general, you either go mobile first or desktop first, with mobile first being the most common because something like 80% of websurfing is done on mobile.

But this page is built desktop-first, which makes sense since you'll mostly be browsing it while writing lean on your desktop machine. So let the default be for desktop and override the things you need to override for mobile.

jakobvase: Looked it up. 60% :)

david-christiansen: OK, that's easy to change.

Out of curiosity, why not? I think it makes it easier to figure out which parts are for which while reading the CSS code. Is there a drawback other than it not being so common?

jakobvase: Yeah I'm curious too. I know that 1/2 pixels are a thing on retina screens, so in this specific case you would get neither option at 700.5 px, but you can 'solve' that with `min-width: 700.01px`, though that's not satisfying.

For me I think of the stylesheet as being for the most-used platform, with tweaks to make it pretty on other platforms too. It reveals what the author of the sheet was thinking. If you make changes to your 'main' platform, you probably want most of those changes on the other platforms too (colors, font styles, the feel of the website). But I'm just theorizing. I also think there is value in 'just doing it like we always do', especially if you're likely to hire another consultant for this in the future.

PR #264, opened 2025-01-06
fix: use declared syntax for all Verso syntax

### LINKED ISSUE #257: Antiquotes in `block` category do not work
Minimized to be verso-free:
```lean
import Lean

open Lean.Parser (rawIdent)

declare_syntax_cat arg_val
syntax (name:=arg_str) str : arg_val

declare_syntax_cat argument
syntax (name:=anon) arg_val : argument
syntax (name:=named) ident ":=" arg_val : argument

declare_syntax_cat inline

declare_syntax_cat block
syntax (name:=directive) "directive{" rawIdent argument* "}" "[" block* "]": block

syntax (name:=header) inline* : block

example : Lean.TSyntax `block → Bool :=
  fun
  | `(block| directive{foo $args*}[$blocks]) => true
  | _ => false
```
which fails with "invalid 'many' parser combinator application, parser did not consume anything".

Changing
```diff
-syntax (name:=header) inline* : block
+syntax (name:=header) inline+ : block
```
makes the problem goes away.

david-christiansen:
All of the syntax declarations used for Verso are now used for pattern-matching in the main elaborator, which is reason to believe that they're now reasonably comprehensive. This should make it much easier for others to use them to write Verso extensions.

Closes #257


src/verso/Verso/Syntax.lean
eric-wieser, line 1: It would be super helpful to me if this file could have a module docstring explaining the that this is internal syntax for writing syntax transformations, and that the user-facing syntax is defined in `Elsewhere.lean`.

david-christiansen: Sure thing - how's the one I just pushed?

src/verso/Verso/Syntax.lean
```diff
@@ -6,6 +6,20 @@ Author: David Thrane Christiansen
 
 import Verso.SyntaxUtils
 
+/-!
+
+This file contains an internal syntax that's used to represent Verso documents. It's primarily
+intended for use in quasiquotations that match against the output of the Verso parser, though it
+does serve to document the expected format of the syntax data produced by the parser. A later step
+transforms the syntax in this module into the actual syntax of Verso documents, potentially invoking
```
eric-wieser: At first read this sounds backwards; are you saying that the syntax _parsers_ here are automatically transformed into parsers for verso documents? Or that there is a delaborator that converts the syntax here into the document syntax? Or did you mean that the document syntax is transformed into this internal syntax?

src/verso/Verso/Syntax.lean
eric-wieser: Looks great, though perhaps worth a `# Heading` line

src/verso/Verso/Syntax.lean
david-christiansen: The latter - I'll clarify.

src/verso/Verso/Syntax.lean
david-christiansen: If you have a concrete suggestion for a header that you'd find helpful, I'm happy to incorporate it

PR #286, opened 2025-02-06
chore: bump to new nightly

david-christiansen:
This fixes compatibility with upstream changes to LSP internals. In the process, the refactoring of the syntax representation was propagated to the semantic token highlighter, greatly improving it.


david-christiansen:
This should unblock you, @nomeata, once I get it propagated down to the manual itself.

PR #301, opened 2025-02-16
feat: improve legibility

jakobvase:
The intention here is to improve legibility of the reference manual and other texts that will be generated from verso.

> feat: reduce font size on desktop
> 
> Since we use `rem` almost everywhere, the effect on the whole page is
> that it is slightly zoomed out. It doesn't increase the number of
> characters per line.
> 
> 16px is taken from MDN, the rust book, and stack overflow. It is the
> most common font size on the web.

> feat: increase line-height for body content
> 
> I investigated line heights from MDN and the rust book, and picked the
> lesser one out of caution. To me, it increases legibility.


david-christiansen:
This is great, thank you!

PR #310, opened 2025-02-22
feat: several style changes for verso manuals

jakobvase:
This has a bunch of ideas related to issues https://github.com/leanprover/reference-manual/issues/316 , https://github.com/leanprover/reference-manual/issues/317 , https://github.com/leanprover/reference-manual/issues/318 , and https://github.com/leanprover/reference-manual/issues/319 .

It builds on top of the sans-serif pull request, so this is the combined changes.

Items addressed:

   - Add a fixed header
   - Reduce font size on desktop
   - Increase line height for body content
   - Make margins on all sides of definition boxes the same size.
   - Decrease the font size of the definition header text
   - Reduce index header font size
   - Make the index header not stick on mobile
   - Make the index always single-column
   - Disable toc hide/show on desktop

I'll add more tomorrow.

### DIFF
--- a/src/verso-manual/VersoManual/Docstring.lean
+++ b/src/verso-manual/VersoManual/Docstring.lean
@@ -458,9 +458,16 @@ def docstringStyle := r#"
 }
 
 .namedocs .text {
+  /* Causes margins on child elements to collapse inside the element, such
+     that the margins don't extend into parent with the background color.
+     The effect is that weird borders in the definition box don't happen anymore. */
+  display: flow-root;
   background-color: white;
-  padding: 1.5rem;
-  margin-top: 0.5rem;
+  /* If there is no text, still show a white box. */
+  min-height: 3rem;
+  /* Add a padding. this is the same as the margin applied to the first and last child.
+     The effect is that the padding looks the same size on all sides. */
+  padding: 0 1.5rem;
 }
 
 .namedocs .text > pre {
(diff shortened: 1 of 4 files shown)


jakobvase:
Added more:

- fix missing footer buttons after adding the header
- when jumping to an id, display it below the header
- when jumping to an id in the index, display it below both headers
- add back ToC transition on mobile (it works this time)
- don't store opened/closed toc items in the browser

The last two items are related. Storing UI state in the browser comes with a bunch of gotchas that you need to handle, and I don't think the functionality is worth it. For example: A user opens the reference manual on Basic types/Strings, decides he wants to see the full toc, but then just goes to the next page. The root toc is stored as open, but the basic types is not stored at all. Now, from the user's perspective as she navigates around the manual, for some reason the root ToC stays open, but the basic types open and close based on what she's viewing.

I opted to remove the stateful navigation and instead just implement a rule that the bottom two levels of navigation are always open. This is consistent and doesn't get in the way of navigating.

david-christiansen:
Is this ready for review/merge now? That was my understanding, but the draft bit is still set.

jakobvase:
Ready! The one in the language reference needs to move the pointer to Verso before that's ready, I'll update it when this is merged.

david-christiansen:
The font size and the line height are already in `main`, but it followed the usual Lean dev practice of squash-merging with PR granularity.

The changes here are independent enough that I plan to rebase-merge anyway.

Mind if I do a surgical rebase on this PR branch to drop the first two commits in order to keep the history making sense?

david-christiansen:
I'll just do it, but I'll back up the original state just in case :)

david-christiansen:
I'm seeing an issue that I think is a genuine bug. The header is an empty bar on the default manual config:

image

This makes sense - no quick jump bar upstream yet, and no logo configured.

I would think that in this situation, either there should be no header, or there should be something in it.

The default font should probably also be sans here, but that's something I can just fix myself.

david-christiansen:
OK, I pushed a quick kludge to fix this in 12bbdef:
```
:root:has(header:empty) {
    --verso-header-height: 0px;
}
```

When you put the title up in the header, this should probably go away.

PR #312, opened 2025-03-03
chore: bump toolchain to v4.17.0

jcommelin:


kim-em:
This was attempting to move the lean-toolchain to an earlier version, which wasn't going to work. Instead I've directly tagged an earlier commit as `v4.17.0`, and pushed that tag.

PR #315, opened 2025-03-08
Put the title in the header for manuals built on Verso

jakobvase:
This PR is in draft, because it needs the title of the manual currently being worked on - right now it is hardcoded to "The Lean Language Reference". Other than that, it's ready to go.

The move to the ToC is implemented such that it is easy to overwrite for
dependents. They will need to add the following css to their theme
file:

```css
/* Move the title from the header to the toc when there is not enough room. */
@media screen and (max-width: 1200px) {
  .toc-title {
    display: block;
  }

  .header-title {
    display: none;
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,24 @@ try to make the system more extensible.
 
 To generate the Verso documentation for Verso itself, run `generate.sh`.
 
+### Customization of manuals built with Verso
+
+To change the point at which the title moves from the header bar to
+the ToC, add something like the following css to your theme file.
+
+```css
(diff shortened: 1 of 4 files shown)


david-christiansen:
I fixed the TODOs.

The blank header space workaround when there's no logo seems to be broken now, and my CSS skills aren't sufficient for a quick selector tweak:
image

Any suggestions?

david-christiansen:
OK, I found a way to hide the header again (when the title is hidden, the logo div is empty, and there's exactly two children). It's fragile, but it'll do for now.

PR #325, opened 2025-03-16
fix: prev-next buttons always take up exactly half the screen

jakobvase:
This fixes two issues:

- The first and last pages of the book, the next/previous button would take up the full width, which was too much.
- On mobile, when flipping through the pages, sometimes the "next" button would move to below the "previous" button, which meant that I accidentally went back a page instead of forwards. Now, their location is consistent.

The drawback is that there will now, on mobile, be more pages where the next/prev buttons span multiple lines. But I think the tradeoff is worth it.


david-christiansen:
Thanks!

PR #327, opened 2025-03-16
fix: code block was overflowing

jakobvase:
Before:

Screen Shot 2025-03-16 at 13 33 10

After:

Screen Shot 2025-03-16 at 13 33 14


david-christiansen:
Thank you!

PR #329, opened 2025-03-16
feat: new design for definition boxes

jakobvase:
Draft because I need feedback on the design, because I need to move the permalink on mobile, and because I need to fix cases where there is no title (eg `/Type-Classes/Instance-Declarations`).

Change the design of the definition boxes. It's lighter and has more room for text.

Screenshot 2025-03-16 at 15 30 14
Screenshot 2025-03-16 at 15 29 13

### DIFF
--- a/src/verso-manual/VersoManual/Docstring.lean
+++ b/src/verso-manual/VersoManual/Docstring.lean
@@ -454,12 +454,9 @@ instance [BEq α] [Hashable α] [ToJson α] : ToJson (HashSet α) where
 def docstringStyle := r#"
 .namedocs {
   position: relative;
-  border: solid 2px #99b3c0;
-  background-color: #99b3c0;
-  padding-left: 1px;
-  padding-right: 1px;
-  padding-bottom: 1px;
-  padding-top: 1.5rem;
+  border-left: solid 3px #d2dfe5;
+  background-color: #d2dfe5;
+  padding-top: .75rem;
   margin-bottom: 1rem;
 }
 
@@ -471,7 +468,7 @@ def docstringStyle := r#"
   background-color: white;
   /* Add a padding. this is the same as the margin applied to the first and last child.
      The effect is that the padding looks the same size on all sides. */
-  padding: 0 1.5rem;
+  padding: 0 1rem;
 }
 
 .namedocs .text > pre {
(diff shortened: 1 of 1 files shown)


jakobvase:
Screenshot 2025-03-16 at 15 35 51

jakobvase:
Screenshot 2025-03-16 at 15 34 58

david-christiansen:
Looks nicer!

One thing I'm a bit unsure about with this design is the labels that say what kind of thing is documented. In the old design, they're off to the side and clearly not part of the syntax being shown. In this new one, being on the line right above, there's some risk of confusing them.

For most, but not all, the current label text matches the keyword used to declare the name in question. If they're seen as keywords, then we have two problems: when they aren't keywords, it'll be discongruous. When they do look like the keywords, then it ends up looking like bad syntax highlighting.

I think there's basically two good possibilities:
 1. Redo the presentation so they're clearly not keywords. This could entail moving the label or changing the text so that fewer of the labels match Lean syntax ("structure" must match, but `def` could be "definition"). This lets us keep some of the finer distinctions made here (e.g. "inductive predicate" vs "inductive type") that may be useful for some newer users.
 2. Delete the labels and use keywords, syntax-highlighting them as appropriate. This is nice because it matches Lean, but not all documented things have clear keywords that introduce them. This would require a rethink of how these boxes are used for tactics, syntax, compiler options, and Lake subcommands.

I'm partial to number 1, because I think it helps have a consistent communication of "here's the reference for this kind of thing". Can we do something with the layout/typography to more clearly distinguish the label from the signature/title?

jakobvase:
> One thing I'm a bit unsure about with this design is the labels that say what kind of thing is documented. In the old design, they're off to the side and clearly not part of the syntax being shown. In this new one, being on the line right above, there's some risk of confusing them.

I've been confused about the relative importance of these labels, and it's something I've been meaning to ask about. How important are they? Should they only be visible on hover? Would it be OK on mobile if the label covered some of the name of the thing being described, or is that a no-go?

> I'm partial to number 1, because I think it helps have a consistent communication of "here's the reference for this kind of thing". Can we do something with the layout/typography to more clearly distinguish the label from the signature/title?

I agree.

PR #338, opened 2025-03-22
fix: several small UI fixes

jakobvase:
fix: display the label consistently over namedocs headers

fix: overflowing code examples, LawfulMonad

I think this is only displayed in a few places, one of them being
`LawfulMonad.{u, v}`. Would it be possible to apply some formatting to
those examples too, so the scrolling becomes unnecessary?

fix: darker warning color (and centralize it)

I picked #E7A71D, as it has 2.11 contrast to white, which should be OK
for visual cues. The previous color was #efd871, which only has 1.42.
See https://webaim.org/resources/contrastchecker/?fcolor=EFD871&bcolor=FFFFFF.

### DIFF
--- a/src/verso-manual/VersoManual/Docstring.lean
+++ b/src/verso-manual/VersoManual/Docstring.lean
@@ -486,6 +486,7 @@ def docstringStyle := r#"
 }
 
 .namedocs .label {
+  display: block;
   font-size: smaller;
   font-family: var(--verso-structure-font-family);
   position: absolute;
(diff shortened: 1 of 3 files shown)


david-christiansen:
> I think this is only displayed in a few places, one of them being
LawfulMonad.{u, v}. Would it be possible to apply some formatting to
those examples too, so the scrolling becomes unnecessary?

Yes, for sure. This formatting uses the Lean pretty printer, and I'm not sure why it's not inserting line breaks there. I'll see if there's anything quick to do, otherwise I can address it after some higher-priority items.

PR #382, opened 2025-05-02
README: fix example output path

mgeier:
This seems to have been copy-pasted from the previous paragraph ...


david-christiansen:
Thanks!

PR #383, opened 2025-05-02
README: fix HTML directory for http.server

mgeier:
I'm not sure if I did something wrong, but when I ran the previous command, I only got a `http-multi` sub-directory, I didn't get `http-single`.


david-christiansen:
Thank you!

PR #389, opened 2025-05-10
fix: behavior of SVG tags inside the HTML

algebraic-dev:


src/verso/Verso/Output/Html.lean
```diff
@@ -110,7 +115,22 @@ private def mustClose : List String :=
    "h2", "h3", "h4", "h5", "h6", "hgroup", "header", "footer", "address", "pre",
    "blockquote", "ol", "ul", "menu", "dl", "figure", "figcaption", "main", "search",
    "div", "a", "em", "strong", "small", "s", "cite", "q", "dfn", "abbr", "ruby",
-   "data", "time", "code", "var", "samp", "kbd", "sub", "sup", "i"]
+   "data", "time", "code", "var", "samp", "kbd", "sub", "sup", "i",
+   "altGlyph", "a", "altGlyphDef", "altGlyphItem", "animate", "animateColor",
```
david-christiansen: I think it would be nice to keep the SVG elements separate from the HTML ones, to help us out case we ever need to make the set configurable (e.g. due to some other format with contradictory requirements). 
```suggestion
   "data", "time", "code", "var", "samp", "kbd", "sub", "sup", "i"] ++
  -- SVG tags begin here
  [ "altGlyph", "a", "altGlyphDef", "altGlyphItem", "animate", "animateColor",
```

david-christiansen: I think my additional concern is that there may be some further incompatibility between HTML and XML that comes up here, where we'd really end up needing two different output routines rather than having one that works for both. In that case, it'd be nice to be able to quickly move the two lists of element names around.

PR #393, opened 2025-05-13
fix: behavior of SVG tags inside the HTML

algebraic-dev:
Reopening the PR after rewriting the git history


david-christiansen:
Thanks!

Issue #436, opened 2025-06-13
More Bibliography type


SnO2WMaN:
There are few bibliograpy type in now `Article`, `Thesis`, `ArXiv`, `InProceedings` in Verso Manual, however it is useful (or necessary) for adding more types e.g. `Book`, `Web` or else.

PR #443, opened 2025-06-19
feat: Add `Book` for bibliography

SnO2WMaN:
related #436

### DIFF
--- a/src/verso-manual/VersoManual/Bibliography.lean
+++ b/src/verso-manual/VersoManual/Bibliography.lean
@@ -59,11 +59,9 @@ structure Book where
   year : Int
   url : Option String := none
   isbn : Option String := none -- TODO: make concrete type `ISBN10` or `ISBN13`
-  publisher : Option (Doc.Inline Manual) := none
+  publisher : Doc.Inline Manual
   edition : Option (Doc.Inline Manual) := none
-  series : Option (Doc.Inline Manual) := none
-  number: Option Int := none
-
+  series : Option ((Doc.Inline Manual) × Int) := none
 deriving ToJson, FromJson, BEq, Hashable, Ord
 
 structure ArXiv where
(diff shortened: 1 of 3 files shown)


src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -167,6 +186,13 @@ def Citable.bibHtml (go : Doc.Inline Genre.Manual → HtmlT Manual (ReaderT Exte
   | .arXiv p =>
     let authors ← andList <$> p.authors.mapM go
     return {{ {{authors}} s!", {p.year}. " {{ link {{"“" {{← go p.title}} "”"}} }} ". arXiv:" {{p.id}} }}
+  | .book p =>
+    let authors ← andList <$> p.authors.mapM go
+    return {{
+      {{authors}} s!", {p.year}. " {{ link {{"“" {{← go p.title}} "”"}} }} ". "
+      {{ p.series.map (λ (s, v) => {{ "Volume " <strong>s!"{v}"</strong> " of " /- TODO: insert series -/ ", "}}) |>.getD .empty }}
```
SnO2WMaN, line 193: I assume it's rare for a book reference to include only series or volume without the other, so I’ve treated them as a pair. However, I’m unsure how to represent this pairing properly in the citation formatting (manupulating `Html` or etc. is too complex).

src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -53,6 +53,17 @@ structure Thesis where
...
 
+structure Book where
+  title : Doc.Inline Manual
+  authors : Array (Doc.Inline Manual)
+  year : Int
+  url : Option String := none
+  isbn : Option String := none -- TODO: make concrete type `ISBN10` or `ISBN13`
+  publisher : Doc.Inline Manual
+  edition : Option (Doc.Inline Manual) := none
+  series : Option ((Doc.Inline Manual) × Int) := none
```
david-christiansen, line 64: I think this misunderstands "volume" in citations.

I would use volume for a single work that's published in multiple physical books, e.g. _The Art of Computer Programming_ by Knuth. A series is usually a collection of disparate books by multiple authors created over a period of time, and it's IMHO less important here.

That is, if I'm citing TAoCP, it matters a lot whether it's vol. 1 or 3. But knowing that a book is part of LNCS is less important.

What about:
```suggestion
  /-- Volume in a multi-volume work -/
  volume : Option Nat
  /-- Series name and position within it, e.g. "Great Lectures in Mathematics 47"
  series : Option ((Doc.Inline Manual) × Int) := none
```

src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -77,6 +88,7 @@ inductive Citable where
   | thesis : Thesis → Citable
   | arXiv : ArXiv → Citable
   | article : Article → Citable
+  | book : Book → Citable
```
david-christiansen, line 91: This is fine, but I do feel like this datatype is getting out of hand and it will soon be time to find an open representation here rather than the current one. A type class, perhaps.

david-christiansen: (this is not something for you to do, unless you're inspired)

src/verso-manual/VersoManual/Bibliography.lean
david-christiansen: See comment on datatype

david-christiansen:
Thanks!

src/verso-manual/VersoManual/Bibliography.lean
SnO2WMaN: I don't have great opinion, so if it is more suitable (like bibtex), I'll change.

SnO2WMaN:
Sorry for lating to reply for reviewing. I'll fix

SnO2WMaN:
I can't spare time to fix this issue so I closed this PR.

jcreedcmu:
Closing for inactivity for now.

PR #444, opened 2025-06-20
feat: allow parts to disable showing inline HTML ToC

jrr6:
This PR adds the ability for a `Part` to disable showing an inline table of contents on a rendered HTML page with an `htmlToc` option. This is necessary for error explanations, since we will render a custom table of explanations with additional metadata rather than using Verso's built-in table of contents.


david-christiansen:
Thanks!

PR #456, opened 2025-06-25
feat: allow hiding Lean block proof states in blog genre

david-christiansen:


david-christiansen:
@algebraic-dev Does this meet your needs?

david-christiansen:
@algebraic-dev OK, it should be on all the code blocks you plan to use now. Please let me know if anything else is needed!

PR #472, opened 2025-07-21
feat: allow customising blog header

b-mehta:
Currently, any page generated by the blog genre starts with the magic string "<!DOCTYPE html>". This PR modifies that to, by default, use `Html.doctype` (as used by the manual genre). 
It also allows customising that string to something else, which is useful to replace it with a Jekyll-friendly front-matter. The default behaviour is left unchanged.

### DIFF
--- a/src/verso-blog/VersoBlog.lean
+++ b/src/verso-blog/VersoBlog.lean
@@ -800,7 +800,8 @@ private def filterString (p : Char → Bool) (str : String) : String := Id.run <
 
 open Template in
 def blogMain (theme : Theme) (site : Site) (relativizeUrls := true) (linkTargets : Code.LinkTargets := {})
-    (options : List String) (components : Components := by exact %registered_components) :
+    (options : List String) (components : Components := by exact %registered_components)
+    (header : String := Html.doctype) :
     IO UInt32 := do
   let hasError ← IO.mkRef false
   let logError msg := do hasError.set true; IO.eprintln msg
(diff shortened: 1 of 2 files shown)


src/verso-blog/VersoBlog.lean
```diff
@@ -819,7 +820,7 @@ def blogMain (theme : Theme) (site : Site) (relativizeUrls := true) (linkTargets
     linkTargets := linkTargets,
     components := components
   }
-  let (((), st), _) ← site.generate theme initGenCtx .empty {}
+  let (((), st), _) ← site.generate theme initGenCtx .empty {} (header := header)
```
david-christiansen: Perhaps the header should be part of the `GenerateM` reader monad context, rather than a parameter that's passed about here? Having it as an optional parameter to lots of functions seems to be a chance to forget one of them and get tricky bugs, and it is something that we'd like to supply at the start and then propagate downwards, right?

david-christiansen:
I think this is a great way to accomplish this. I just have a small API quibble (in a comment).

Thank you very much!

src/verso-blog/VersoBlog.lean
b-mehta: That makes sense, but doesn't that mean the user needs to make their own `blogMain` function, since the current one doesn't allow giving Generate.Context as an input? `cfg` currently is built out of `opts` below, so I suppose the options are either to amend that, or pass `header` as an input to `blogMain` but insert it into the `Generate.Context` on line 813 and have it part of the reader context internally. Perhaps the second is cleaner?

david-christiansen:
Rebasing on `main` should fix the CI issue

src/verso-blog/VersoBlog.lean
b-mehta: I've done the second

david-christiansen:
Thanks!

PR #476, opened 2025-07-23
fix: parser errors getting silently lost

david-christiansen:


david-christiansen:
This still needs testing on the big corpus of Verso documents out there before a merge

PR #478, opened 2025-07-24
fix: mobile Safari rendering issue

david-christiansen:


david-christiansen:
This will be merged after verifying that this really fixes it in the manual.

PR #480, opened 2025-07-28
chore: fix logError interpolation in inlineHtml

b-mehta:
The string wasn't being interpolated, so error messages were for "{x}" rather than for `x` itself.


david-christiansen:
Thanks!

PR #514, opened 2025-08-22
fix: don't insert newlines when rendering manual inline names

### LINKED ISSUE #513: Spurious newline following code inlines in TeX output of Manual genre
**To reproduce**:

Generate the pdf of the verso manual itself using `./generate.sh` as usual.
In `_out/tex/main.pdf`, observe (around line 1596 at time of writing)
```
...
In the document, they consist of instances of \Verb|Manual.Block|
 and \Verb|Manual.Inline|
, respectively:
...
```
This leads to the pdf admitting a line break between `Manual.Inline` and the comma, and in other cases,
when a code inline occurs at the end of a line of Verso input, it creates a double newline in the TeX output, leading to an undesired paragraph break, for example
```
The \Verb|mode|

determines whether it is rendered in inline mode or display mode; even display-mode math is an
``` 
(around line 1071 of main.tex)

**Cause**:

Discussed with @david-christiansen and it looks like `inline_extension Inline.name` is taking all the children of a name inline, which will in practice be a singleton list of a code inline, and emit them with newlines appended. It should simply concatenate them without newlines. This is easier to express now that `TeX` has a `Coe` instance from `Array TeX`, which it didn't at the time the code was written.

jcreedcmu:
Remove newline insertion from inline name rendering. Addresses part of #513.

### DIFF
--- a/src/verso-manual/VersoManual/InlineLean.lean
+++ b/src/verso-manual/VersoManual/InlineLean.lean
@@ -744,10 +744,7 @@ where
 inline_extension Inline.name where
   traverse _ _ _ := do
     pure none
-  toTeX :=
-    some <| fun go _ _ content => do
-      pure <| .seq <| ← content.mapM fun b => do
-        pure <| .seq #[← go b, .raw "\n"]
+  toTeX := some <| fun go _ _ content => content.mapM go
   extraCss := [highlightingStyle]
   extraJs := [highlightingJs]
   extraJsFiles := hlJsDeps


jcreedcmu:
This resolves the `\Verb|Manual.Inline|` newline but not the `\Verb|mode|` newline mentioned in the issue. I don't know why yet. Still looking.

jcreedcmu:
The extraneous `\Verb|mode|` newline is inside a docstring and is maybe happening for other reasons, e.g. markdown processing might be introducing it. I don't see any reason not to land this PR as an incremental step in the direction of correctness, but I'll keep looking into how else TeX output can be made coherent.

PR #516, opened 2025-08-23
chore: migrate parser tests to files

david-christiansen:
Previously, the parser tests were all inline `#guard_msgs` commands. While this gave quick feedback, it made it impossible to test part of the parser while tests were failing and it increased the parser's elaboration time. Additionally, updating them all was very difficult, requiring clicking on them one at a time.

This PR moves the tests to a Lake test runner, with most tests in files that can be batch-updated.


david-christiansen:
This will also facilitate upstreaming, as this for of test is more compatible with Lean's test quite.

PR #518, opened 2025-08-26
fix: generate placeholder text for missing `toTex` implementations

kpadmasola:
Generate placeholder text for missing `toTex` implementations instead failing.

### DIFF
--- a/src/verso-manual/VersoManual/Basic.lean
+++ b/src/verso-manual/VersoManual/Basic.lean
@@ -1271,7 +1271,8 @@ instance : TeX.GenreTeX Manual (ReaderT ExtensionImpls IO) where
     let some descr := (← readThe ExtensionImpls).getBlock? b.name
       | panic! s!"Unknown block {b.name} while rendering.\n\nKnown blocks: {(← readThe ExtensionImpls).blockDescrs.toArray |>.map (·.fst) |>.qsort (·.toString < ·.toString)}"
     let some impl := descr.toTeX
-      | return .command "textcolor" #[] #[ .raw "red", .raw s!"Missing toTeX for Block {b.name}" ]
+      | TeX.logError s!"Block {b.name} doesn't support TeX"
+        return \TeX{\textcolor{"red"}{s!"Missing toTeX for Block {b.name}"}}
     impl goI goB id b.data content
   inline go i content := do
     let some id := i.id
@@ -1279,7 +1280,8 @@ instance : TeX.GenreTeX Manual (ReaderT ExtensionImpls IO) where
     let some descr := (← readThe ExtensionImpls).getInline? i.name
       | panic! s!"Unknown inline {i.name} while rendering.\n\nKnown inlines: {(← readThe ExtensionImpls).inlineDescrs.toArray |>.map (·.fst) |>.qsort (·.toString < ·.toString)}"
     let some impl := descr.toTeX
-      | return .command "textcolor" #[] #[ .raw "red", .raw s!"Missing toTeX for Inline {i.name}" ]
+      | TeX.logError s!"Inline {i.name} doesn't support TeX"
+        return \TeX{\textcolor{"red"}{s!"Missing toTeX for Inline {i.name}"}}
     impl go id i.data content
 
 
--- a/src/verso-manual/VersoManual/Basic.lean
+++ b/src/verso-manual/VersoManual/Basic.lean
@@ -1271,8 +1271,9 @@ instance : TeX.GenreTeX Manual (ReaderT ExtensionImpls IO) where
     let some descr := (← readThe ExtensionImpls).getBlock? b.name
       | panic! s!"Unknown block {b.name} while rendering.\n\nKnown blocks: {(← readThe ExtensionImpls).blockDescrs.toArray |>.map (·.fst) |>.qsort (·.toString < ·.toString)}"
     let some impl := descr.toTeX
-      | TeX.logError s!"Block {b.name} doesn't support TeX"
-        return .empty
+      | --TeX.logError s!"Block {b.name} doesn't support TeX"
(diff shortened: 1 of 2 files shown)


kpadmasola:
@david-christiansen With this change, the `main.tex` gets generated. However, running `lualatex main.tex` fails with errors. I have categorized the errors below along with the fixes to enable successful conversion to `main.pdf`.


- FIX: Need to add the following for showing error text in red color
	`\usepackage{xcolor}`
	`\colorlet{RED}{black}`
- FIX: Replace `\Verb` with `\verb`
- ERROR: Cannot have verbatim text inside section titles
    - `LaTeX Error: \verb illegal in argument.`
    - e.g., `\subsubsection{The \verb|ToFormat| Class}`
    - FIX: We may need something like https://tex.stackexchange.com/questions/152229/strange-error-related-to-verb-in-section-titles
    - Or https://tex.stackexchange.com/questions/24574/how-to-put-verb-command-inside-of-textbf-block
    - Or https://tex.stackexchange.com/questions/2790/when-should-one-use-verb-and-when-texttt
- ERROR: `LaTeX Error: \verb ended by end of line`  — happens in hyperlink args
    - e.g., in `\hyperlink{"https://lean-lang.org/doc/reference/4.23.0-rc2/find/?domain=Verso.Genre.Manual.section&name=do-notation"}{\verb|do|`
    -notation}.Because the resulting value is treated as a side-effect-free term, the compiler may re-order,`
    - FIX: use `\texttt` instead
- ERROR: `LaTeX Error: Command \item invalid in math mode.`
    - e.g., `.\item The syntax \verb|have (eq := h) pat := e|`
		 `is equivalent to \verb|match h : e with | pat => _|`
    - FIX: In the second line there is a `|` inside `\verb`
    - So we should use some other delimiter character, such as `!`
    - Or use `\begin{verbatim}`
    - Also `\item` should probably begin on a new line, perhaps?
- ERROR: `LaTeX Error: Lonely \item--perhaps a missing list environment.`
    - Issue due to `\hyperlink` second arg having underscore
    - FIX: escape underscore
- ERROR: `! Text line contains an invalid character.`
    * e.g., `l.32508 ... or. Usually accessed via the \verb|^^^`
    * FIX: `Try \verb|\^\^\^|`
- ERROR: `LaTeX Error: \verb illegal in argument.`
    - FIX: Replace with `\texttt` instead

src/verso-manual/VersoManual/Basic.lean
```diff
@@ -1271,17 +1271,15 @@ instance : TeX.GenreTeX Manual (ReaderT ExtensionImpls IO) where
     let some descr := (← readThe ExtensionImpls).getBlock? b.name
       | panic! s!"Unknown block {b.name} while rendering.\n\nKnown blocks: {(← readThe ExtensionImpls).blockDescrs.toArray |>.map (·.fst) |>.qsort (·.toString < ·.toString)}"
     let some impl := descr.toTeX
-      | TeX.logError s!"Block {b.name} doesn't support TeX"
```
david-christiansen: I think this error message should be retained. It's important to prominently tell authors that their document isn't as they expect, without them needing to proofread the resulting text.

src/verso-manual/VersoManual/Basic.lean
```diff
@@ -1271,17 +1271,15 @@ instance : TeX.GenreTeX Manual (ReaderT ExtensionImpls IO) where
...
-        return .empty
+      | return .command "textcolor" #[] #[ .raw "red", .raw s!"Missing toTeX for Block {b.name}" ]
     impl goI goB id b.data content
   inline go i content := do
     let some id := i.id
       | panic! "Inline {i.name} wasn't assigned an ID during traversal"
     let some descr := (← readThe ExtensionImpls).getInline? i.name
       | panic! s!"Unknown inline {i.name} while rendering.\n\nKnown inlines: {(← readThe ExtensionImpls).inlineDescrs.toArray |>.map (·.fst) |>.qsort (·.toString < ·.toString)}"
     let some impl := descr.toTeX
-      | TeX.logError s!"Inline {i.name} doesn't support TeX"
```
david-christiansen: Similarly, I think this error message should be retained too. It's good to emit better TeX though.

src/verso/Verso/Doc/TeX.lean
```diff
@@ -41,7 +41,7 @@ def logError [Monad m] (message : String) : TeXT g m Unit := do
 def header [Monad m] (name : TeX) : TeXT g m TeX := do
   let opts ← options
   let some i := opts.headerLevel
-    | logError s!"No more header nesting available at {name.asString}"; return \TeX{\textbf{\Lean{name}}}
```
david-christiansen: This should also be a proper error, and it should insert a notice like the other two into the document that is unmistakable.

david-christiansen (changes requested):
This is a good start! Thanks!

But I think it's important that authors don't have to proofread their document to know if something went wrong. The best situation is that they get error messages and a non-zero return code, but also output that is as readable as possible.

david-christiansen:
Thanks for the big list of things that need fixing! That'll be really helpful to check off as it goes.

src/verso-manual/VersoManual/Basic.lean
kpadmasola: done.

src/verso-manual/VersoManual/Basic.lean
kpadmasola: done.

src/verso/Verso/Doc/TeX.lean
kpadmasola: done.

david-christiansen:
I'm away from the computer for a few days but will look again as soon as I'm back. Thanks again!

PR #519, opened 2025-08-28
chore: Add an integration test for TeX output

jcreedcmu:
Not yet as happy as I could be with this. It might be ok to land as incremental forward progress.

Flaws:
- The currently passing test actually encodes undesired behavior; the spurious newline from https://github.com/leanprover/verso/issues/513 shows up. Ideally I'd rather not check in a red test, nor a test that's green but semantically failing.
- WriterT seem to exist in mathlib not the standard prelude, so I'm using StateT
- Doesn't yet support the golden-test pattern as I'd like, because I still need to figure out how to get a string elaborated.
- Maybe TeX generation could be more of a pure function? Need to investigate more. It might be that it's too intertwined with config and IO-monadic logging to conveniently untangle, in which case I might not bother.

### DIFF
--- a/src/tests/TestMain.lean
+++ b/src/tests/TestMain.lean
@@ -4,6 +4,8 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Author: David Thrane Christiansen
 -/
 
+import Verso
+import VersoManual
 import VersoSearch.PorterStemmer
 import Tests
 
@@ -31,13 +33,37 @@ def testStemmer (_ : Config) : IO Unit := do
 
 def testParser (dir : System.FilePath) (fn : Lean.Parser.ParserFn) : Config → IO Unit := fun config =>
   Verso.GoldenTest.runTests {
-    testDir := ("src/tests/parser" : System.FilePath) / dir,
+    testDir := "src/tests/parser" / dir,
     runTest := fn.test,
     updateExpected := config.updateExpected
   }
 
+/--
+Tests manual-genre TeX generation. `dir` is a subdirectory specific to a particular test document.
+`doc` is the document itself.
+-/
+def testTexOutput (dir : System.FilePath) (doc : Verso.Doc.Part Verso.Genre.Manual) :
+    Config →  IO Unit := fun config =>
+  let versoConfig : Verso.Genre.Manual.Config := {
+    destination := "src/tests/integration" / dir / "output",
+    emitTeX := true,
+    emitHtmlMulti := false
+  }
+
+  let runTest : IO Unit  :=
+    open Verso Genre Manual in do
+    let logError (msg : String) := IO.eprintln msg
+    ReaderT.run (emitTeX logError versoConfig doc) extension_impls%
+
+  Verso.Integration.runTests {
+    testDir := "src/tests/integration" / dir,
+    updateExpected := config.updateExpected,
+    runTest
+  }
+
 open Lean.Parser in
 open Verso.Parser in
+open Verso.Integration in
 def tests := [
   testStemmer,
   testParser "metadataBlock" metadataBlock,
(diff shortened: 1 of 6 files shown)


src/tests/Tests/Integration/SampleDoc.lean
```diff
@@ -0,0 +1,42 @@
...
+And then here's a paragraph break.
+-/
+def sample_constant := Unit
+
+#docs (Manual) sample_part "Title of the Doc" :=
+:::::::
+
+%%%
+shortTitle := "ShortTitle"
+authors := ["Harry Q. Bovik"]
```
david-christiansen, line 30: 😆

src/verso-manual/VersoManual.lean
```diff
@@ -328,17 +324,28 @@ def emitTeX (logError : String → IO Unit) (config : Config) (text : Part Manua
       | continue
     packages := packages.insertMany d.usePackages
     preambleItems := preambleItems.insertMany d.preamble
-  withFile (dir.join "main.tex") .write fun h => do
-    if config.verbose then
-      IO.println s!"Saving {dir.join "main.tex"}"
-    h.putStrLn (preamble text.titleString authors date packages.toList preambleItems.toList)
+  let putStrLn (line : String) : StateT String Id Unit := do
+    modify (fun s => s ++ line ++ "\n")
+  let tex : StateT String Id Unit := do
```
david-christiansen: `StateT s Id` is also known as `StateM s` in Lean

src/verso-manual/VersoManual.lean
```diff
@@ -328,17 +324,28 @@ def emitTeX (logError : String → IO Unit) (config : Config) (text : Part Manua
...
-    h.putStrLn postamble
+      putStrLn c.asString
+    putStrLn postamble
+  pure (StateT.run tex "").2
+
+open IO.FS in
+def emitTeX (logError : String → IO Unit) (config : Config) (text : Part Manual) : ReaderT ExtensionImpls IO Unit := do
+  let dir := config.destination.join "tex"
+  ensureDir dir
+  let tex ← makeTeX logError config text
```
david-christiansen: Rather than making this a string, what if we passed an `IO.FS.Stream` here, potentially via a Reader? That can be hooked up to a file handle, but it can also be set up to log the output. That way, we don't commit ourselves to creating a string in memory for all TeX output, but we can still test the output by doing so.

src/tests/TestMain.lean
```diff
@@ -37,9 +37,16 @@ def testParser (dir : System.FilePath) (fn : Lean.Parser.ParserFn) : Config →
     updateExpected := config.updateExpected
   }
 
+def testManualTeX (dir : System.FilePath) : Config → IO Unit := fun config =>
+  Verso.GoldenTest.runTests {
+    testDir := ("src/tests/manual-tex" : System.FilePath) / dir,
+    runTest := fun _s => Verso.ManualTexTest.run -- TODO(admit multiple files)
```
david-christiansen: It seems to me that we should have a different test runner for this. The one used for the parser is really intended for friendlier testing of pure functions, while here I think we really want to check the whole run. What if this test were instead one that ran a command in a directory and then compared a given list of files to their expected versions? Then we could use the build command for the document unmodified, making it a better integration test, and the test running code could stay simpler and more to the point.

david-christiansen:
Thanks! Comments inline.

I think it's OK to encode _some_ behavior that we don't want in a test in order to capture behavior that we do want, so long as this situation is clearly documented so someone fixing the unwanted behavior doesn't get confused.

jcreedcmu:
I've written a test runner that does a somewhat generic "run an IO.unit given a directory, and compare the whole set of files produced to a parallel expected directory", as suggested. Error handling for that isn't the greatest yet.

One definite omission is [the fact that it doesn't actually handle `--update-expected`](https://github.com/leanprover/verso/pull/519/files#diff-c301715386e498c41c05553bd3001113d7bc65d19f6f2dbb62b2c5f6395f82cdR52) correctly yet. I'm still poking around to see the right way to `cp -r`, bearing in mind that the right way may simply be `.spawn`ing `cp -r`.

jcreedcmu:
This is now based on top of https://github.com/leanprover/verso/pull/543 so don't necessarily invest time in reviewing this closely until the question of whether that should be landed is dealt with.

PR #520, opened 2025-08-30
doc: small clarification on extraCss docstrings

b-mehta:
My reading of these docstrings was that they insert extra css as a string directly, rather than being intended to be file names, so this PR amends the docstrings to explicitly describe css files (in particular because extraCssFiles could also have been taking in file names!)

Please feel free to suggest alternate wording here!

### DIFF
--- a/src/verso-manual/VersoManual.lean
+++ b/src/verso-manual/VersoManual.lean
@@ -174,9 +174,9 @@ structure Config where
   emitHtmlMulti : Bool := true
   wordCount : Option System.FilePath := none
   extraFiles : List (System.FilePath × String) := []
-  /-- Extra CSS to be included inline into every `<head>` -/
+  /-- Extra CSS files to be included inline into every `<head>` -/
   extraCss : List String := []
-  /-- Extra JS to be included inline into every `<head>` -/
+  /-- Extra JS files to be included inline into every `<head>` -/
   extraJs : List StaticJsFile := []
   /-- Extra CSS to be written to the filesystem in the Verso data directory and loaded by each `<head>` -/
   extraCssFiles : Array (String × String) := #[]
--- a/src/verso-manual/VersoManual/Basic.lean
+++ b/src/verso-manual/VersoManual/Basic.lean
@@ -583,7 +583,7 @@ structure InlineDescr where
   -/
   toHtml : Option (InlineToHtml Manual (ReaderT ExtensionImpls IO))
   /--
-  Extra JavaScript to add to a `<script>` tag in the generated HTML's `<head>`
+  Extra JavaScript files to add to a `<script>` tag in the generated HTML's `<head>`
   -/
   extraJs : List String := []
   /--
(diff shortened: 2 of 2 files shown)


src/verso-manual/VersoManual.lean
```diff
@@ -174,9 +174,9 @@ structure Config where
   emitHtmlMulti : Bool := true
   wordCount : Option System.FilePath := none
   extraFiles : List (System.FilePath × String) := []
-  /-- Extra CSS to be included inline into every `<head>` -/
+  /-- Extra CSS files to be included inline into every `<head>` -/
```
david-christiansen, line 177:
```suggestion
  /-- Extra CSS content to be included inline into every `<head>` -/
```
I think the reference to "files" here is likely to confuse more than it clarifies, because I'd expect it to be a file name.

Or instead, perhaps:
```suggestion
  /-- Extra CSS code to be included inline into every `<head>` -/
```

Would these have clarified it for you?

b-mehta: It's meant to be a file name, isn't it?

b-mehta: Hm, looking at other uses it looks like I misunderstood; I'm not sure why it didn't work in my case.

b-mehta: I tested my example again, and indeed for me, putting anything in the extraCss bit of the config causes it to go into some `<link rel="stylesheet" href="`, which definitely indicates a file name to me.

robsimmons:
I'll talk to @david-christiansen about this tomorrow but I'm pretty sure what's going on here is that the docs are right and the implementation is wrong.

I made a minimal-ish example [here](https://github.com/robsimmons/ovenchar-2025-09-16/blob/main/Main.lean#L12-L13)

```lean4
def config : Config where
   extraCss := ["html { background-color: blue; }"]
   extraCssFiles := #[("foo.css", "html { color: red; }")]
```

and can confirm that the generated `_out/html-multi/index.html` contains

```html
    <link rel="stylesheet" href="html { background-color: blue; }">
    <link rel="stylesheet" href="-verso-data/foo.css">
```

where the former is obviously not correct. However, where verso is used in the Lean reference manual, for example [here](https://github.com/leanprover/reference-manual/blob/main/Manual/Meta/ElanCmd.lean#L258-L269), generated HTML is putting the extraCss, as one would anticipate, in `<style>...</style>` brackets in the head of pages.

If I'm correct, the resolution tomorrow will probably be to close this ticket but file a bug to make the implementation actually work as indicated by the docs. Alternatively, David's suggested "code" change may be reasonable to really drive home unabigiously the incorrectness of the behavior that @b-mehta and I were seeing in our examples.

robsimmons:
Closing this pull request in favor of actually planning to fix it via #529 — thank you @b-mehta for bringing this discrepancy between implementation and documentation to our attention! When I fix #529 I'll probably add something like the "Extra CSS" -> "Extra CSS code" change proposed here by @david-christiansen.

PR #527, opened 2025-09-15
chore: update lychee dependency in github action

robsimmons:
There's a probably-not-important-for-us dependabot update on our lychee version, but the [breaking changes](https://github.com/lycheeverse/lychee-action/releases/tag/v2.0.0) for v2 all seemed to be about sensible changes to the defaults, so it doesn't seem like anything would hold us back from just updating to a v2 dependency


david-christiansen:
CI passes so I'm sure it's fine :-)

PR #530, opened 2025-09-18
fix: correctly handle extraJs/Css in manual genre (closes #529)

### LINKED ISSUE #529: extraCss sometimes shows up as an href instead of correctly in <style> tags
This issue was raised by @b-mehta indirectly via #520 — that PR attempts to fix the spec, but the spec is right and the implementation is wrong; extraCss in a site `Config` is getting included as an href instead of as an inline style between `<style>...</style>` tags.

robsimmons:
- make extraJs a string
- correctly initialize the traverse state with the config
- don't add extraCss to the file list

### DIFF
--- a/src/verso-manual/VersoManual.lean
+++ b/src/verso-manual/VersoManual.lean
@@ -177,7 +177,7 @@ structure Config where
   /-- Extra CSS to be included inline into every `<head>` -/
   extraCss : List String := []
   /-- Extra JS to be included inline into every `<head>` -/
-  extraJs : List StaticJsFile := []
+  extraJs : List String := []
   /-- Extra CSS to be written to the filesystem in the Verso data directory and loaded by each `<head>` -/
   extraCssFiles : Array (String × String) := #[]
   /-- Extra JS to be written to the filesystem in the Verso data directory and loaded by each `<head>` -/
@@ -265,6 +265,8 @@ def traverse (logError : String → IO Unit) (text : Part Manual) (config : Conf
   let remoteContent ← updateRemotes false config.remoteConfigFile (if config.verbose then IO.println else fun _ => pure ())
   let mut state : Manual.TraverseState := {
     licenseInfo := .ofList config.licenseInfo,
+    extraCss := Std.HashSet.emptyWithCapacity |>.insertMany config.extraCss
+    extraJs := Std.HashSet.emptyWithCapacity |>.insertMany config.extraJs
     extraCssFiles := config.extraCssFiles,
     extraJsFiles := config.extraJsFiles,
     remoteContent
@@ -410,7 +412,6 @@ def page (toc : List Html.Toc)
   }
   let extraJsFiles :=
     sortJs <|
-      config.extraJs.toArray.map (true, ·) ++
       state.extraJsFiles.map (false, ·.toStaticJsFile)
   let extraJsFiles := extraJsFiles.map fun
     | (true, f) => (f.filename, f.defer)
@@ -423,9 +424,9 @@ def page (toc : List Html.Toc)
     (repoLink := config.sourceLink)
     (issueLink := config.issueLink)
     (localItems := localItems)
-    -- The extra CSS and JS in the config is not take here because it's used to initialize the
+    -- The extraCss and extraJs in the config is not take here because it's used to initialize the
     -- traverse state and is thus already present.
-    (extraStylesheets := config.extraCss ++ state.extraCssFiles.toList.map ("/-verso-data/" ++ ·.1))
+    (extraStylesheets := state.extraCssFiles.toList.map ("/-verso-data/" ++ ·.1))
     (extraJsFiles := extraJsFiles)
     (extraHead := config.extraHead)
     (extraContents := config.extraContents)


src/verso-manual/VersoManual.lean
```diff
@@ -265,6 +265,8 @@ def traverse (logError : String → IO Unit) (text : Part Manual) (config : Conf
   let remoteContent ← updateRemotes false config.remoteConfigFile (if config.verbose then IO.println else fun _ => pure ())
   let mut state : Manual.TraverseState := {
     licenseInfo := .ofList config.licenseInfo,
+    extraCss := Std.HashSet.emptyWithCapacity |>.insertMany config.extraCss
+    extraJs := Std.HashSet.emptyWithCapacity |>.insertMany config.extraJs
```
robsimmons: I'm guessing there's a prettier way to do this?

david-christiansen: I'd usually write the empty hash set with `{}`.

src/verso-manual/VersoManual.lean
```diff
@@ -410,22 +412,21 @@ def page (toc : List Html.Toc)
...
   let extraJsFiles :=
     sortJs <|
-      config.extraJs.toArray.map (true, ·) ++
       state.extraJsFiles.map (false, ·.toStaticJsFile)
   let extraJsFiles := extraJsFiles.map fun
     | (true, f) => (f.filename, f.defer)
     | (false, f) => ("/-verso-data/" ++ f.filename, f.defer)
   Html.page toc path textTitle htmlBookTitle contents
+    -- The extraCss, extraJs, extraCssFiles, and extraJsFiles in the config are absent here
+    -- because they arere included in the traverse state when it is initialized
```
david-christiansen:
```suggestion
    -- because they are included in the traverse state when it is initialized
```

src/verso-manual/VersoManual.lean
```diff
@@ -174,10 +174,10 @@ structure Config where
   emitHtmlMulti : Bool := true
   wordCount : Option System.FilePath := none
   extraFiles : List (System.FilePath × String) := []
-  /-- Extra CSS to be included inline into every `<head>` -/
+  /-- Extra CSS to be included inline into every `<head>` via `<script>` tags -/
   extraCss : List String := []
-  /-- Extra JS to be included inline into every `<head>` -/
-  extraJs : List StaticJsFile := []
+  /-- Extra JS to be included inline into every `<head>` via `<style>` tags -/
+  extraJs : List String := []
```
david-christiansen, line 180: I think that we still need a config field for this value. This is used for including JS files from the static files directory in each page (e.g. we used it for the search bar when it was still a local plugin in the manual and hadn't been upstreamed yet).

The docstrings should say that a `StaticJsFile` is identified by its name but loaded from the static files directories, while a `JsFile` has a name and contents and is written to a Verso-specific helper directory.

robsimmons: I'm gonna push back on this from a You're Not Gonna Need It perspective — these fields were added for comparison with the `BlockDescr` and `InlineDescr` which only has the String and JsFile versions. 

Directly introducing `StaticJsFile` links can *mostly* be done by just dumping JS into some static directory adding `extraHead` without much loss (you would lose the ability to sort files by `after` dependencies, I think that's it). That feels extremely corner case, especially given that we are presumably going to want to move all this configuration around (from Manual to a more generic HTML genre) at some point.

src/verso-manual/VersoManual.lean
robsimmons: My next attempt is 

```lean4
    extraCss := .insertMany {} config.extraJs
    extraJs := .insertMany {} config.extraJs
```

do you have any ideas to golf that any better?

david-christiansen: Not off the top of my head, if there's no `ofList` defined.

src/verso-manual/VersoManual.lean
david-christiansen: Sounds fine to me.

But I don't think "HTML" is a genre - HTML is an output format, and many different genres will need to produce it. Or are you thinking more that this should be part of a library of "how to configure the HTML output for your genre, assuming it's fairly standard"?

david-christiansen (approved):

david-christiansen:
Assuming you've tested that the reference manual builds correctly with this, please merge!

src/verso-manual/VersoManual.lean
robsimmons: good terminology correction, but yes — most any genre outputting to HTML likely wants a lot of the functionality currently specific to the Manual genre

robsimmons:
@david-christiansen the plot, perhaps, thickens; the reference manual _was_ using the config's `extraCss` as a list of files, which we both decided was wrong 😭.

But we can just move those declarations over to being extraHead, which was the solution I expected above; this is compatible with current verso so I'll want to merge leanprover/reference-manual#595 before this PR.

src/verso-manual/VersoManual.lean
```diff
@@ -174,10 +174,10 @@ structure Config where
   emitHtmlMulti : Bool := true
   wordCount : Option System.FilePath := none
   extraFiles : List (System.FilePath × String) := []
-  /-- Extra CSS to be included inline into every `<head>` -/
+  /-- Extra CSS to be included inline into every `<head>` via `<script>` tags -/
```
lua-vr, line 177: I think the end of the two comments are swapped: here it should be ``via `<style>` tags``, and `<script>` should be below?

src/verso-manual/VersoManual.lean
```diff
@@ -265,6 +265,8 @@ def traverse (logError : String → IO Unit) (text : Part Manual) (config : Conf
   let remoteContent ← updateRemotes false config.remoteConfigFile (if config.verbose then IO.println else fun _ => pure ())
   let mut state : Manual.TraverseState := {
     licenseInfo := .ofList config.licenseInfo,
+    extraCss := .insertMany {} config.extraJs
```
lua-vr, line 268: I believe here it should be `extraCss := .insertMany {} config.extraCss` instead of `extraJs`? I'm trying to use it but it seems `extraCss` does not work without this change.

david-christiansen: It absolutely should, but from what I can see this bug was fixed a long time ago and isn't present today. Can you try again on the latest Verso?

src/verso-manual/VersoManual.lean
david-christiansen: I agree, but this is also fixed already, right?

src/verso-manual/VersoManual.lean
david-christiansen: In any case, https://github.com/leanprover/verso/pull/660 should do a lot to prevent a recurrence.

lua-vr: @david-christiansen, oh, I'm really sorry! I started from the `verso-templates` repository and it didn't occur to me that the package was pinned to `v4.24.0` in the Lakefile. I should have checked before.

david-christiansen: No problem - that repo is now updated, and future updates are both automated and on my release checklist. Thanks!

PR #539, opened 2025-09-29
fix: correct markdown header processing

jcreedcmu:
Previously, parts generated by `addPartFromMarkdown` were at wrong nesting depth for nontrivial structures. Error explanations (for which this code was originally written) did not exercise this but because they have fairly simple structure, without many children of the same header, or decreases of header depth.

This PR simplifies the verso-nesting-depth-to-markdown-nesting-depth state, to a list of `Nat` rather than a list of pair of `Nat`s. I think without sacrificing correctness.

### DIFF
--- a/src/verso-manual/VersoManual/Markdown.lean
+++ b/src/verso-manual/VersoManual/Markdown.lean
@@ -258,15 +258,15 @@ where
 open Verso.Doc.Elab
 
 /--
-Close all sections that have a markdown header level that is ≥
-{name}`level`, to prepare the state for pushing new a part at level
-{name}`level`.
+Close all sections that have a markdown header level that is greater
+than or equal to {name}`level`, to prepare the state for pushing new a
+part at level {name}`level`.
 
 We close a frame in the `.partContext` of `PartElabM.State` exactly in lockstep
 with dropping the head of `inHeaders` in `MDState`.
 -/
 private partial def closeSections {m} [Monad m]
-    [MonadStateOf PartElabM.State m]
+    [MonadError m] [MonadStateOf PartElabM.State m]
     (level : Nat) : MDT m b i Unit := do
   let hdrs := (← getThe MDState).inHeaders
   match hdrs with
@@ -274,12 +274,11 @@ private partial def closeSections {m} [Monad m]
   | docLevel :: more =>
     if docLevel ≥ level then
       -- `default` here because markdown parser provides no source position
-      if let some ctxt' := (← getThe PartElabM.State).partContext.close default then
-        modifyThe PartElabM.State fun st => {st with partContext := ctxt'}
-        modifyThe MDState ({· with inHeaders := more})
-        closeSections level
-      else
-        panic! "tried to close part but couldn't"
+      let some ctxt' := (← getThe PartElabM.State).partContext.close default
+        |  throwError m!"Failed to close verso part corresponding to markdown section: no parts left"
+      modifyThe PartElabM.State fun st => {st with partContext := ctxt'}
+      modifyThe MDState ({· with inHeaders := more})
+      closeSections level
 
 /--
 In our header mapping bookkeeping, create a new section with a new markdown header with level {name}`level`.
@@ -338,3 +337,60 @@ def addPartFromMarkdown {m} [Monad m]
   let ctxt := {headerHandlers := ⟨handleHeaders⟩, elabInlineCode, elabBlockCode}
   let (_, { inHeaders }) ← (addPartFromMarkdownAux md |>.run ctxt |>.run {inHeaders := currentHeaderLevels})
   return inHeaders
+
+/--
+Given a finished part (which may contain subparts) render just the titles as markdown
+headings with a number of # that reflects their nesting depth. Meant to be used
+for debugging/testing only.
+-/
+def displayPartStructure (part : Verso.Doc.Elab.FinishedPart) (level : Nat := 1) : String := match part with
+  | .mk _ _ title _ _ subParts _ =>
+       let partsStr : String := subParts.map (displayPartStructure · (level + 1))
+         |>.toList |> String.join
+       let pref := List.replicate level "#" |> String.join
+       s!"{pref} {title}\n{partsStr}"
+  | .included name => s!"included {name}\n"
+
+/--
+Given a markdown string, parse it and return the displayed part structure.
+-/
+def testAddPartFromMarkdown (input : String) : Elab.TermElabM String := do
+  let some parsed := MD4Lean.parse input
+    | throwError m!"Couldn't parse markdown {input}"
+  let addParts : PartElabM Unit := do
+    let mut levels := []
+    for block in parsed.blocks do
+      levels ← addPartFromMarkdown block levels
+    closePartsUntil 0 0
+  let (_, _, part) ← addParts.run (Syntax.node .none identKind #[]) (mkConst ``Manual) default default
+  part.partContext.priorParts.toList.map displayPartStructure |> String.join |> pure
+
+/--
+info:
+# header1
+## header2-a
+### header3-aa
+## header 2-b
+### header3-ba
+### header3-bb
+#### header4-bba
+### header3-bc
+# another header
+## one more
+-/
+#guard_msgs in
+/- Exercise how inconsistent markdown header nesting depth
+is heuristically fixed. -/
+#eval do
+  IO.println <| "\n" ++ (← testAddPartFromMarkdown r#"
+# header1
+## header2-a
+### header3-aa
+## header 2-b
+##### header3-ba
+#### header3-bb
+###### header4-bba
+### header3-bc
+# another header
+## one more
+"#)
--- a/src/verso-manual/VersoManual/Markdown.lean
+++ b/src/verso-manual/VersoManual/Markdown.lean
@@ -47,9 +47,14 @@ def attr' (val : Array AttrText) : Except String String := do
   | .error e => .error e
   | .ok s => pure s
 
+/-- A mapping from markdown document header levels to actual verso nesting levels.
+The reason we need this is that we make a best effort to repair non-consecutive
+markdown header levels in a reasonable way, so we need to remember which mapping
+choices we already made. -/
+private abbrev HeaderMapping := List (Nat × Nat)
+
 private structure MDState where
-  /-- A mapping from document header levels to actual nesting levels -/
-  inHeaders : List (Nat × Nat) := []
+  inHeaders : HeaderMapping := []
 deriving Inhabited
 
 private abbrev MDT m block inline α := ReaderT (MDContext m block inline) (StateT MDState m) α
@@ -241,7 +246,9 @@ where
 open Verso.Doc.Elab
 
 /--
-Updates the active sections given a new header with `level`.
+Updates the active sections given a new header with `level`. We specifically close
+all sections that have a markdown header level that is ≥ {name}`level`, to prepare
+the state for pushing new a part at level {name}`level`.
 -/
 private partial def closeSections {m} [Monad m]
     [MonadStateOf PartElabM.State m]
@@ -249,15 +256,23 @@ private partial def closeSections {m} [Monad m]
   let hdrs := (← getThe MDState).inHeaders
   match hdrs with
   | [] => modifyThe MDState ({· with inHeaders := [(level, 0)]})
-  | (docLevel, nesting) :: more =>
-    if level ≤ docLevel then
+  | (docLevel, _) :: more =>
+    if docLevel ≥ level then
       if let some ctxt' := (← getThe PartElabM.State).partContext.close default then -- Markdown parser provides no source position
         modifyThe PartElabM.State fun st => {st with partContext := ctxt'}
-        closeSections level
-      if level < docLevel then
         modifyThe MDState ({· with inHeaders := more})
-    else
-      modifyThe MDState ({· with inHeaders := (level, nesting + 1) :: hdrs})
+        closeSections level
+      else
+        panic! "tried to close part but couldn't"
+
+private partial def newSection {m} [Monad m]
+    [MonadStateOf PartElabM.State m]
+    (level : Nat) : MDT m b i Unit := do
+  let hdr := (← getThe MDState).inHeaders
+  let nextNesting := match hdr with
+  | [] => 0
+  | (_, nesting) :: _ => nesting + 1
+  modifyThe MDState ({· with inHeaders := (level, nextNesting) :: hdr})
 
 private partial def addPartFromMarkdownAux {m} [Monad m]
     [MonadLiftT PartElabM m] [MonadStateOf PartElabM.State m]
@@ -265,6 +280,7 @@ private partial def addPartFromMarkdownAux {m} [Monad m]
     : MD4Lean.Block → MDT m Term Term Unit
   | .header level txt => do
     closeSections level
+    newSection level
     let txtStxs ← txt.mapM inlineFromMarkdown |>.run' none
     let titleTexts ← match txt.mapM stringFromMarkdownText with
       | .ok t => pure t
@@ -298,10 +314,10 @@ def addPartFromMarkdown {m} [Monad m]
     [MonadLiftT PartElabM m] [MonadStateOf PartElabM.State m]
     [MonadQuotation m] [AddMessageContext m] [MonadError m]
     (md : MD4Lean.Block)
-    (currentHeaderLevels : List (Nat × Nat) := [])
+    (currentHeaderLevels : HeaderMapping := [])
     (handleHeaders : List (Array Term → m Term) := [])
     (elabInlineCode : Option (Option String → String → m Term) := none)
-    (elabBlockCode : Option (Option String → Option String → String → m Term) := none) : m (List (Nat × Nat)) := do
+    (elabBlockCode : Option (Option String → Option String → String → m Term) := none) : m HeaderMapping := do
   let ctxt := {headerHandlers := ⟨handleHeaders⟩, elabInlineCode, elabBlockCode}
   let (_, { inHeaders }) ← (addPartFromMarkdownAux md |>.run ctxt |>.run {inHeaders := currentHeaderLevels})
   return inHeaders
--- a/src/verso-manual/VersoManual/Markdown.lean
+++ b/src/verso-manual/VersoManual/Markdown.lean
@@ -48,10 +48,18 @@ def attr' (val : Array AttrText) : Except String String := do
   | .ok s => pure s
 
 /-- A mapping from markdown document header levels to actual verso nesting levels.
-The reason we need this is that we make a best effort to repair non-consecutive
-markdown header levels in a reasonable way, so we need to remember which mapping
-choices we already made. -/
-private abbrev HeaderMapping := List (Nat × Nat)
+The list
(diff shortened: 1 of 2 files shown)


jcreedcmu:
I added some incidental comments to some Verso code as well based on my current understanding of it, let me know if I said anything wrong.

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -47,9 +47,26 @@ def attr' (val : Array AttrText) : Except String String := do
   | .error e => .error e
   | .ok s => pure s
 
+/--
+A mapping from markdown document header levels to actual verso nesting levels.
```
david-christiansen:
```suggestion
A mapping from Markdown document header levels to actual Verso nesting levels.
```
(proper nouns)

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -47,9 +47,26 @@ def attr' (val : Array AttrText) : Except String String := do
...
+The values in the list are markdown header levels. Their position in the list
```
david-christiansen:
```suggestion
The values in the list are Markdown header levels. Their position in the list
```

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -47,9 +47,26 @@ def attr' (val : Array AttrText) : Except String String := do
...
+For example, the list
+    `[5,4,2,1]`
+is understood as associating:
+- markdown level 1 to verso nesting 0
+- markdown level 2 to verso nesting 1
+- markdown level 4 to verso nesting 2
+- markdown level 5 to verso nesting 3
+
+We need to keep this state to appropriately repair non-consecutive
+markdown header levels.
```
david-christiansen:
```suggestion
is the Verso nesting level, with the final element being Verso level 0.
For example, the list
    `[5,4,2,1]`
is understood as associating:
- Markdown level 1 to Verso nesting 0
- Markdown level 2 to Verso nesting 1
- Markdown level 4 to Verso nesting 2
- Markdown level 5 to Verso nesting 3

We need to keep this state to appropriately repair non-consecutive
Markdown header levels.
```

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -47,9 +47,26 @@ def attr' (val : Array AttrText) : Except String String := do
...
+-/
+public def HeaderMapping := List Nat
```
david-christiansen, line 65: Why is this a `def` rather than an `abbrev`?

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -47,9 +47,26 @@ def attr' (val : Array AttrText) : Except String String := do
...
+deriving Inhabited
+
 private structure MDState where
-  /-- A mapping from document header levels to actual nesting levels -/
-  inHeaders : List (Nat × Nat) := []
+  inHeaders : HeaderMapping := []
```
david-christiansen, line 69: This seems like a nice simplification.

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -241,23 +258,37 @@ where
 open Verso.Doc.Elab
 
 /--
-Updates the active sections given a new header with `level`.
+Close all sections that have a markdown header level that is greater
+than or equal to {name}`level`, to prepare the state for pushing new a
+part at level {name}`level`.
+
+We close a frame in the `.partContext` of `PartElabM.State` exactly in lockstep
+with dropping the head of `inHeaders` in `MDState`.
```
david-christiansen:
```suggestion
Closes all sections that have a Markdown header level that is greater
than or equal to {name}`level`, to prepare the state for pushing new a
part at level {name}`level`.

We close a frame in the {name (full:=PartElabM.State.partContext)}`partContext` of {name}`PartElabM.State` exactly in lockstep
with dropping the head of {name (full:=MDState.inHeaders)}`inHeaders` in {name}`MDState`.
```

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -241,23 +258,37 @@ where
...
-        modifyThe PartElabM.State fun st => {st with partContext := ctxt'}
-        closeSections level
-      if level < docLevel then
-        modifyThe MDState ({· with inHeaders := more})
-    else
-      modifyThe MDState ({· with inHeaders := (level, nesting + 1) :: hdrs})
+  | [] => pure ()
+  | docLevel :: more =>
+    if docLevel ≥ level then
+      -- `default` here because markdown parser provides no source position
```
david-christiansen:
```suggestion
      -- `default` here because the Markdown parser provides no source position
```

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Hm, there may not be any need to have it be a `def` anymore. I think I may have done this during development to make sure nothing unexpected was happening with typeclass inference. On some level, I'd like to express that this type is an implementation choice and that callers outside the module should not know what it is defined as, but it's not clear to me that `public def` even enforces that. Could switch it to `abbrev`.

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -241,23 +258,37 @@ where
...
+      let some ctxt' := (← getThe PartElabM.State).partContext.close default
+        |  throwError m!"Failed to close verso part corresponding to markdown section: no parts left"
+      modifyThe PartElabM.State fun st => {st with partContext := ctxt'}
+      modifyThe MDState ({· with inHeaders := more})
+      closeSections level
+
+/--
+In our header mapping bookkeeping, create a new section with a new markdown header with level {name}`level`.
```
david-christiansen:
```suggestion
In our header mapping bookkeeping, creates a new section with a new Markdown header with level {name}`level`.
```
Lean docstrings should use the indicative mood, rather than the imperative. (Other languages are split on this point, we decided to go with the Rust convention)

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -241,23 +258,37 @@ where
...
+This must be accompanied by pushing a new part.
```
david-christiansen: Should this be just one function that does both things, to reduce the potential for mistakes with this API?

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -270,6 +301,7 @@ private partial def addPartFromMarkdownAux {m} [Monad m]
       | .ok t => pure t
       | .error e => throwError m!"Unsupported Markdown in header:\n{e}"
     let titleText := titleTexts.foldl (· ++ ·) ""
+    newSection level
     PartElabM.push {
```
david-christiansen: It would be nice if this were something like:
```
startMarkdownSection level {
```

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Yeah. I do hope I was sufficiently careful in ensuring it's correct. I did look and couldn't find any place where the invariant *wasn't* already maintained that the now-absent `Nat` always corresponds to the distance from the end of the list.

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Great, good to know, I was operating under the influence of https://leanprover-community.github.io/contribute/commit.html which indeed specifies the opposite convention for commit messages. Nothing wrong with different conventions applying in different contexts.

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Sure, I considered doing exactly that. Will do.

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Agree.

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
-    (elabBlockCode : Option (Option String → Option String → String → m Term) := none) : m (List (Nat × Nat)) := do
+    (elabBlockCode : Option (Option String → Option String → String → m Term) := none) : m HeaderMapping := do
   let ctxt := {headerHandlers := ⟨handleHeaders⟩, elabInlineCode, elabBlockCode}
   let (_, { inHeaders }) ← (addPartFromMarkdownAux md |>.run ctxt |>.run {inHeaders := currentHeaderLevels})
   return inHeaders
+
+/--
+Given a finished part (which may contain subparts) render just the titles as markdown
+headings with a number of # that reflects their nesting depth. Meant to be used
+for debugging/testing only.
```
david-christiansen:
```suggestion
Renders the entire structure of a finished part as Mardown-style headings, with a
number of `'#'s` that reflects their nesting depth. This is a tool for debugging/testing
only.
```
This sounds very useful!

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
+-/
+def displayPartStructure (part : Verso.Doc.Elab.FinishedPart) (level : Nat := 1) : String := match part with
```
david-christiansen:
```suggestion
open Verso.Doc.Elab in
def displayPartStructure (part : FinishedPart) (level : Nat := 1) : String := match part with
```
For a super long line like this, I think it's also nice to break the line before the `match`.

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
+  | .mk _ _ title _ _ subParts _ =>
+       let partsStr : String := subParts.map (displayPartStructure · (level + 1))
+         |>.toList |> String.join
+       let pref := List.replicate level "#" |> String.join
```
david-christiansen:
```suggestion
       let pref := "".pushn '#' level
```
But we do this so often that we should probably name it.

Are levels here 0-indexed or 1-indexed? Should this have one more `'#'`?

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
+       s!"{pref} {title}\n{partsStr}"
+  | .included name => s!"included {name}\n"
+
+/--
+Given a markdown string, parse it and return the displayed part structure.
```
david-christiansen:
```suggestion
Parses a Markdown string, returning the displayed part structure.
```

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
+-/
+def testAddPartFromMarkdown (input : String) : Elab.TermElabM String := do
```
david-christiansen, line 362: Given that we're in `TermElabM` anyway, what about rendering this to a set of nested trace messages? Would that be easier to work with interactively? Or is it just not worth the effort?

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
+### header3-bb
+#### header4-bba
+### header3-bc
+# another header
+## one more
+-/
+#guard_msgs in
+/- Exercise how inconsistent markdown header nesting depth
+is heuristically fixed. -/
+#eval do
```
david-christiansen, line 389: 🎉

src/verso-manual/VersoManual/Markdown.lean
```diff
@@ -298,10 +330,67 @@ def addPartFromMarkdown {m} [Monad m]
...
+## header 2-b
+### header3-ba
+### header3-bb
+#### header4-bba
+### header3-bc
+# another header
+## one more
+-/
+#guard_msgs in
+/- Exercise how inconsistent markdown header nesting depth
```
david-christiansen:
```suggestion
/- Exercises how inconsistent Markdown header nesting depth
```

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -256,6 +256,7 @@ partial def FinishedPart.toTOC : FinishedPart → TOC
     .mk titleString titleStx endPos (subParts.map toTOC)
   | .included name => .included name
 
+/-- Information describing a part still under construction. -/
```
```suggestion
/--
Information describing a part still under construction.

During elaboration, the current position in the document is represented by a stack of these frames, with each frame representing a layer of document section nesting. As the Verso document elaborator encounters new headers, stack frames are pushed and popped as indicated by the header's level.
-/
```

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: I feel I'm following the convention of the [CommonMark Spec](https://spec.commonmark.org/0.31.2/) in that "The heading level is equal to the number of # characters in the opening sequence." and I'm happy to add that citation to the function doc for clarity. Though maybe I'm misunderstanding your intent in asking the question... lmk.

src/verso/Verso/Doc/Elab/Monad.lean
david-christiansen: (please fill this paragraph before merging PR if you accept the suggestion - GitHub comments being in a proportional font makes it tough to do here)

jcreedcmu: Can do.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -264,15 +265,26 @@ structure PartFrame where
   priorParts : Array FinishedPart
```
david-christiansen, line 277:
```suggestion
  /--
  The sibling parts at the same nesting level as the part represented by this frame. These siblings
  are earlier in the document and have the same parent.
  -/
  priorParts : Array FinishedPart
```

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -264,15 +265,26 @@ structure PartFrame where
...
 
+/-- Turn an previously active {name}`PartFrame` into a {name}`FinishedPart`. -/
 def PartFrame.close (fr : PartFrame) (endPos : String.Pos) : FinishedPart :=
   let (titlePreview, titleInlines) := fr.expandedTitle.getD ("<anonymous>", #[])
   .mk fr.titleSyntax titleInlines titlePreview fr.metadata fr.blocks fr.priorParts endPos
 
+/-- Information available while constructing a part. It extends
+{name}`PartFrame` because that data represents the current frame. The
+field {name PartContext.parents}`parents` represents other parts above
+us in the hierarchy that are still being built. -/
```
```suggestion
/--
Information available while constructing a part. It extends {name}`PartFrame`
because that data represents the current frame. The field
{name PartContext.parents}`parents` represents other parts above
us in the hierarchy that are still being built.
-/
```

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -264,15 +265,26 @@ structure PartFrame where
...
 structure PartContext extends PartFrame where
   parents : Array PartFrame
 deriving Repr, Inhabited
 
+/-- The current nesting level is the number of frames in the stack of
+parent parts being built. -/
```
```suggestion
/--
The current nesting level is the number of frames in the stack of parent
parts being built.
-/
```
Our style is to put `/--` and `-/` on the same line only for 1-line docstrings where the entire construction fits in our chosen column limit.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -264,15 +265,26 @@ structure PartFrame where
...
 def PartContext.level (ctxt : PartContext) : Nat := ctxt.parents.size
+
+/-- Close the current part. It is appended to {name}`priorParts`, and
+the top of the stack of our parents becomes the current frame. Returns
+{name}`none` if there are no parents. -/
```
```suggestion
/--
Closes the current part.
The resulting {name}`FinishedPart` is appended to {name}`priorParts`, and
the top of the stack of our parents becomes the current frame. Returns
{name}`none` if there are no parents.
-/
```

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Don't think I'm inclined to for this test in particular, but as a general thing, I am interested in being more familiar with what can be done with trace messages. Where's documentation on that? Nothing obvious comes up searching the reference manual for 'nested trace'.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -284,6 +296,7 @@ def PartContext.close (ctxt : PartContext) (endPos : String.Pos) : Option PartCo
     priorParts := fr.priorParts.push <| ctxt.toPartFrame.close endPos
   }
 
+/-- Make the frame {name}`fr` the current frame. -/
```
```suggestion
/--
Makes the frame {name}`fr` the current frame. The former current frame is saved to the stack.
-/
```

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: is it `withTraceNode` you're thinking of? That's new to me.

src/verso-manual/VersoManual/Markdown.lean
david-christiansen: We're not using the module system yet, so it does not. Definitions with `def` are still definitionally equal to their RHS. If you really want to hide the choice, you'd need a structure with a private constructor, but that'd be way overkill.

I think this is fine as a `def`, so long as there's a reason, which there is.

src/verso-manual/VersoManual/Markdown.lean
david-christiansen: This also matches the implementations of this operation that I had to do as part of docstrings upstream, so I'm pretty sure it's right (or at least that this is a viable model!).

src/verso-manual/VersoManual/Markdown.lean
david-christiansen: No, that's more of an internals thing. I don't think it's well-documented anywhere other than [generated API docs](https://leanprover-community.github.io/mathlib4_docs/Lean/Message.html#Lean.MessageData.trace), and those are quite sparse.

david-christiansen:
Apart from nitpicky docstring stuff, I think this is a nice simplification and I think the docs were all correct.

Did you have `set_option doc.verso true` in the file? If not, then Verso docstrings probably shouldn't be used quite yet :)

david-christiansen:
(or, alternatively, make sure to enable it)

src/verso-manual/VersoManual/Markdown.lean
jcreedcmu: Cool, I'm going to punt on changing this from anything richer than `String`, then, but I'm glad for the suggestion making me find out about more things.

src/verso/Verso/Doc/Elab/Monad.lean
jcreedcmu: Yes, good, I think I correctly perceived that convention and this was a mistake even with respect to my own expectations.

jcreedcmu:
Verso docstrings were later enabled in https://github.com/leanprover/verso/pull/547

PR #540, opened 2025-09-30
refactor: pattern-match destructuring to projection for DocElabContext

robsimmons:
Modify pattern match destructuring to field projection where that's an obvious move. 

This is intended to make it less of a big diff elsewhere where I want to explore adding fields to DocElabContext. I'm also curious if stylistically there's a preferred version — I certainly would be inclined to use the (existing) pattern-destructuring notation if I had less than 3 or so fields and the fields were very stable.


david-christiansen:
I think this is a good change to make.

I tend to prefer the matching style when I think that most use sites are likely to need some kind of review when the datatype changes. If I suspected that all elements of the context were likely to be related, and the addition of a new one meant changing other code, then the red squigglies would be a good opportunity to check them.

In this case, where the fields are not closely semantically related, but instead are bundled because they're used in the same part of the program, I think the field accessor style is indeed better to make it easier to change the contents, as you point out.

PR #541, opened 2025-09-30
fix: ambient level off-by-one error

robsimmons:


jcreedcmu:
Seems right to me. Consider maybe https://github.com/leanprover/verso/pull/542 ?

src/verso/Verso/Doc/Elab.lean
```diff
@@ -334,7 +334,7 @@ partial def _root_.Lean.Doc.Syntax.header.command : PartCommand
     let titleString := headerStxToString (← getEnv) stx
     let ambientLevel ← currentLevel
     let headerLevel := headerLevel.getNat + 1
-    if headerLevel > ambientLevel + 1 then throwErrorAt stx "Wrong header nesting - got {headerLevel} but expected at most {ambientLevel}"
+    if headerLevel > ambientLevel + 1 then throwErrorAt stx "Wrong header nesting - got {headerLevel} but expected at most {ambientLevel + 1}"
```
david-christiansen: An even better improvement would be something like:
```suggestion
    if headerLevel > ambientLevel + 1 then throwErrorAt stx "Wrong header nesting - got {"#".pushn '#' headerLevel} but expected at most {"##".pushn '#' ambientLevel}"
```
(please check that the number of hash marks lines up)

This way users don't have to think about 0-indexed vs 1-indexed header levels.

PR #543, opened 2025-09-30
fix: restore parser tests

jcreedcmu:
This PR is made to find out the reason why the tests were deleted, or, in the event there wasn't a reason, to put them back.

### DIFF
--- a/src/tests/parser/arg/arg_0001.expected
+++ b/src/tests/parser/arg/arg_0001.expected
@@ -1,3 +1,4 @@
 Success! Final stack:
-  (Lean.Doc.Syntax.anon (Lean.Doc.Syntax.arg_ident `x))
+  (Lean.Doc.Syntax.anon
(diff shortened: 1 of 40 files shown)


jcreedcmu:
The several commits in this PR were deliberately constructed so that I could
(a) see the changes induced by `--update-expected` as a diff, and
(b) see that there weren't very many such changes if I preemptively changed `Verso.Syntax` to `Lean.Doc.Syntax` in all of the `.expected` files, enough to manually review with confidence.

david-christiansen:
When the parser was upstreamed, the tests went with it: https://github.com/leanprover/lean4/tree/master/tests/lean/docparse

I don't think there's a reason to have them two places, and the tests should follow the code. I'll go ahead and close this, please feel free to re-open if there's a detail I missed.

jcreedcmu:
Nope, that's the answer, didn't know why I didn't think of that

PR #549, opened 2025-10-03
refactor: prepares for seralizing highlights

robsimmons:
Pulls out common code from the two inline generating functions and the two block generating functions to reduce cutpaste in a first redundanct elimination step, also eliminates some dead code.

Checked that this did not change the sha1sum of the C files generated by compiling the reference manual.

### DIFF
--- a/src/verso-manual/VersoManual/InlineLean.lean
+++ b/src/verso-manual/VersoManual/InlineLean.lean
@@ -261,7 +261,6 @@ def lean : CodeBlockExpanderOf LeanBlockConfig
       if config.show then
         warnLongLines col? str
 where
-  withNewline (str : String) := if str == "" || str.back != '\n' then str ++ "\n" else str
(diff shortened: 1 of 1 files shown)


src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -530,24 +501,9 @@ def inst : RoleExpanderOf LeanBlockConfig
       let hls := (← highlight stx #[] (PersistentArray.empty.push tree))
 
       if config.show then
-        ``(Inline.other (Verso.Genre.Manual.InlineLean.Inline.lean $(quote hls)) #[Inline.code $(quote term.getString)])
+        mkInlineLeanSyntax hls term.getString
```
robsimmons: The only objection I can see to this one is that it's not as aggressive of a refactor as it could have been — the two uses of mkInlineLeanSyntax are heavily cut-pasted

robsimmons:
Replaced by #550

PR #550, opened 2025-10-03
refactor: prepare to serialize highlights

robsimmons:
1. Pulls out common code from the two inline generating functions and the two block generating functions, towards the goal of minimizing cut and paste when we switch from just `$(quote hlt)` to a more complicated serialization procedure for including highlighted material in files.
2. Eliminates some dead code.
3. Fixes a small bug that meant that `-show` didn't work on `lean` or `inst` roles in VersoManual. This is broken on the current release candidate, but I don't think the code path was ever exercised and noone noticed:

   ```lean
   import VersoManual
   open Verso.Genre.Manual.InlineLean
   #doc (Verso.Genre.Manual) "Bonk" =>
   {lean -show}`43 + 2`
   ```

Stress checked that this did not change the sha1sum of the HTML files generated by compiling the reference manual.

Replaces #549

### DIFF
--- a/src/verso-manual/VersoManual/InlineLean.lean
+++ b/src/verso-manual/VersoManual/InlineLean.lean
@@ -172,7 +172,7 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
 De-indents and returns (syntax of) a Block representation containing highlighted Lean code.
 The argument `hls` must be a highlighting of the parsed string `str`.
 -/
-private def mkHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str: StrLit) : DocElabM Term := do
+private def toHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str: StrLit) : DocElabM Term := do
   if !shouldShow then
     return ← ``(Block.concat #[])
 
@@ -189,7 +189,7 @@ private def mkHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str:
 Returns (syntax of) an Inline representation containing highlighted Lean code.
 The argument `hls` must be a highlighting of the parsed string `str`.
 -/
-private def mkHighlightedLeanInline (shouldShow : Bool) (hls : Highlighted) (str : StrLit) : DocElabM Term := do
+private def toHighlightedLeanInline (shouldShow : Bool) (hls : Highlighted) (str : StrLit) : DocElabM Term := do
   if !shouldShow then
     return ← ``(Inline.concat #[])
 
@@ -252,7 +252,7 @@ def lean : CodeBlockExpanderOf LeanBlockConfig
       for cmd in cmds do
         hls := hls ++ (← highlight cmd nonSilentMsgs cmdState.infoState.trees)
 
-      mkHighlightedLeanBlock config.show hls str
+      toHighlightedLeanBlock config.show hls str
     finally
       if !config.keep then
         setEnv origEnv
@@ -360,7 +360,7 @@ def leanTerm : CodeBlockExpanderOf LeanInlineConfig
           logMessage msg
 
       let hls := (← highlight stx #[] (PersistentArray.empty.push tree))
-      mkHighlightedLeanBlock config.show hls str
+      toHighlightedLeanBlock config.show hls str
 
 /--
 Elaborates the provided Lean term in the context of the current Verso module.
@@ -451,7 +451,7 @@ def leanInline : RoleExpanderOf LeanInlineConfig
 
       let hls := (← highlight stx #[] (PersistentArray.empty.push tree))
 
-      mkHighlightedLeanInline config.show hls term
+      toHighlightedLeanInline config.show hls term
 
 
 /--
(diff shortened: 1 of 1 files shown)


src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,34 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
...
 
+/--
+De-indent and return (syntax of) a Block representation containing highlighted Lean code
+The argument `hls` must be a highlighting of the parsed string `str`.
+-/
+private def mkHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str: StrLit) : DocElabM Term := do
+  if !shouldShow then
+    return ← ``(Block.concat #[])
+
+  let col? := (← getRef).getPos? |>.map (← getFileMap).utf8PosToLspPos |>.map (·.character)
```
robsimmons, line 179: This is the only line in the PR that makes me nervous, because col is being read from a lot of stateful stuff that I don't understand and this change moves it from _before_ the elaboration/installation-in-environment process to _after_ that process.

david-christiansen: Stepping through it:

* `(← getRef)` is a `Reader` effect, so no need to worry about mutation. It tracks the "current reference syntax", which is the syntax on which errors/warnings/info should be reported if there's no other more specific place to do it.
* `getPos?` is returning an `Option` that contains the first character position of the current reference syntax
* `(← getFileMap)` uses another `Reader` effect to get the current file map, which is a data structure that stores the byte offsets of all line endings in the current file. This is used to avoid scanning the whole thing while converting between LSP positions (which are line number paired with UTF-16 character index into the line) and Lean string positions. Still no need to worry about mutation. `FileMap.utf8PosToLspPos` converts a Lean string position into an LSP position.

So there's little to fear here. It says "Get the current reference syntax's start position as a UTF-8 offset. If it exists, convert it to an LSP position, then extract the column component."

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,34 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
     if messages.hasErrors then
       throwErrorAt blame "No error expected in code block, one occurred"
 
+/--
+De-indent and return (syntax of) a Block representation containing highlighted Lean code
+The argument `hls` must be a highlighting of the parsed string `str`.
```
david-christiansen: Lean docstrings should be indicative mood:
```suggestion
De-indents and returns (syntax of) a Block representation containing highlighted Lean code.
The argument `hls` must be a highlighting of the parsed string `str`.
```

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,34 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
...
+-/
+private def mkHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str: StrLit) : DocElabM Term := do
```
```suggestion
private def toHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str: StrLit) : DocElabM Term := do
```
Normally we use `mk` for constructors or things like them. How about `to`?

src/verso-manual/VersoManual/InlineLean.lean
david-christiansen: I guess the way this might go wrong is if the reference syntax is different here that it was in the original code - I'll try to keep that in mind.

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,34 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
...
+  | .none => hls
+  | .some col => hls.deIndent col
+
+  let range := Syntax.getRange? str
+  let range := range.map (← getFileMap).utf8RangeToLspRange
+  ``(Block.other (Block.lean $(quote hls) (some $(quote (← getFileName))) $(quote range)) #[Block.code $(quote str.getString)])
+
+/--
+Return (syntax of) an Inline representation containing highlighted Lean code.
+The argument `hls` must be a highlighting of the parsed string `str`.
```
```suggestion
Returns (syntax of) an Inline representation containing highlighted Lean code.
The argument `hls` must be a highlighting of the parsed string `str`.
```

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,34 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
...
+-/
+private def mkHighlightedLeanInline (shouldShow : Bool) (hls : Highlighted) (str : StrLit) : DocElabM Term := do
```
```suggestion
private def toHighlightedLeanInline (shouldShow : Bool) (hls : Highlighted) (str : StrLit) : DocElabM Term := do
```

david-christiansen:
Great stuff, thanks!

src/verso-manual/VersoManual/InlineLean.lean
robsimmons: right - the concern was that in the original code the col? computation was _before_ a bunch of effectful computation and now it's _after_  all that effectual computation. However, the fact that `getRef` is a Reader effect — I was assuming it was an IORef, incorrectly — makes it seem much less likely something went wonky there,

PR #552, opened 2025-10-06
refactor: reduce output C size with redundancy elimination and serialization

robsimmons:
`Highlighted` objects contain highly redundant information, and the result is that the C code we output at present redundantly represents this redundant information.

This PR steals a trick that @david-christiansen introduced in teorth/analysis#352 of exporting `Highlighted` to the existing `ExportCode` format, serializing the resulting data to JSON, and producing syntax that transforms that resulting string back to Highlighted. (Aggressive compiler optimizations could undo this work, but that does not seem to be an issue at present.)

Leads to a 11% reduction in the output size of C code for the reference manual (as reported by `find .lake/build/ir -name "*.c" | xargs du -c | tail -n 1`) — 1049888 to 936912.

### DIFF
--- a/src/verso-manual/VersoManual/InlineLean.lean
+++ b/src/verso-manual/VersoManual/InlineLean.lean
@@ -169,6 +169,15 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
       throwErrorAt blame "No error expected in code block, one occurred"
 
 /--
+Within the DocElabM monad, `← quoteHighlightViaSerialization hls` should result in a Term with the
+same denotation as `quote hls`, but that often produces smaller code because it quotes a compressed
+version of the highlight.
+-/
+private def quoteHighlightViaSerialization (hls : Highlighted) : DocElabM Term := do
+  let repr := hlToExport hls
+  ``(hlFromExport $(quote repr))
+
+/--
 De-indents and returns (syntax of) a Block representation containing highlighted Lean code.
 The argument `hls` must be a highlighting of the parsed string `str`.
 -/
@@ -183,7 +192,7 @@ private def toHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str:
 
   let range := Syntax.getRange? str
   let range := range.map (← getFileMap).utf8RangeToLspRange
-  ``(Block.other (Block.lean $(quote hls) (some $(quote (← getFileName))) $(quote range)) #[Block.code $(quote str.getString)])
+  ``(Block.other (Block.lean $(← quoteHighlightViaSerialization hls) (some $(quote (← getFileName))) $(quote range)) #[Block.code $(quote str.getString)])
 
 /--
 Returns (syntax of) an Inline representation containing highlighted Lean code.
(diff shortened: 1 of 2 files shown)


src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,16 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
     if messages.hasErrors then
       throwErrorAt blame "No error expected in code block, one occurred"
 
+/--
+Produces, within the DocElabM, a Syntax expression that denotes the `hsl` value. Specifically,
```
```suggestion
Produces the syntax of an expression that denotes the `hls` value. Specifically,
```
Docstrings are always paired with signatures, whether in hovers or online docs, so no need to mention the monad here.

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,16 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
...
+within the DocElabM monad, `← quoteHighlightViaSerialization hls` will result in a syntax term that
+represents the same highlight as `quote hls`, but will hopefully produce smaller code because of
+quoting a compressed version of the highlight.
```
```suggestion
quoting a compressed version of the highlighted code.
```
The parameter should probably be called `hl` rather than `hls`, because it's not an array of highlighted code but rather a single one. And this isn't a highlight _per se_, the code comes along for the ride.

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,16 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
     if messages.hasErrors then
       throwErrorAt blame "No error expected in code block, one occurred"
 
+/--
+Produces, within the DocElabM, a Syntax expression that denotes the `hsl` value. Specifically,
+within the DocElabM monad, `← quoteHighlightViaSerialization hls` will result in a syntax term that
```
```suggestion
within the DocElabM monad, `← quoteHighlightViaSerialization hls` will result in a `Term` that
```

src/verso/Verso/Code/Highlighted.lean
```diff
@@ -19,6 +19,33 @@ open Std (HashMap)
...
+-/
+def hlToExport (hl: SubVerso.Highlighting.Highlighted) : String :=
+  hl.exportCode |>.toJson |>.compress
+
+/--
+Recover the `Highlighted` data from its serialization.
+
+Can only be expected to work on the output of `hlToExport`.
+-/
+def hlFromExport (exportLit : String) : SubVerso.Highlighting.Highlighted :=
```
david-christiansen:
```suggestion
def hlFromExport! (exportLit : String) : SubVerso.Highlighting.Highlighted :=
```
The `!` expresses that it will panic on failure (which should also be mentioned in the API docs)

src/verso/Verso/Code/Highlighted.lean
```diff
@@ -19,6 +19,33 @@ open Std (HashMap)
...
 
+
+/--
+Serialize `Highlighted` code as a string.
+
+Nothing should be assumed about the `String` output except that it can be passed to `hlFromExport`
+to recover the original input.
+-/
+def hlToExport (hl: SubVerso.Highlighting.Highlighted) : String :=
+  hl.exportCode |>.toJson |>.compress
```
david-christiansen:
```suggestion
  hl.exportCode.toJson.compress
```

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -168,6 +168,16 @@ def reportMessages {m} [Monad m] [MonadLog m] [MonadError m]
...
+represents the same highlight as `quote hls`, but will hopefully produce smaller code because of
+quoting a compressed version of the highlight.
+-/
+private def quoteHighlightViaSerialization (hls : Highlighted) : DocElabM Term := do
```
david-christiansen, line 177: Can we find a less-long name for this one? Like `quoteCompressed`?

robsimmons: My main argument for the long name would be that there are several approaches to compression... including similar I expect to be exploring this quarter.

david-christiansen: OK :)

Issue #555, opened 2025-10-07
Verso-HTML output has wrong relative paths


jcreedcmu:
This behavior is observable on verso `nightly-testing`.

Steps to reproduce:

1. `lake build LibTarget:literate` for, e.g. some library target `LibTarget`
2. `mkdir outer_dir`
3. `lake exe verso-html .lake/build/literate/ outer_dir/inner_dir`
4. `python3 -m http.server'
5. Browse to http://localhost:8000/outer_dir/inner_dir/LibTarget/

Observe:
- ToC-like links on sidebar do work. (expected)
- Other hyperlinks in content do not work. (unexpected) If their `href` is changed from absolute to relative, they do work. (This fact relies on the `<base>` tag present in the output html, which does establish where the site root is)
- There are a couple 404s for JSON content in console. (unexpected) This is due to `fetch` requests trying to fetch relative to the site root via two independent mechanisms that both try to "go up the appropriate number of directories" and since both are active, we go up *twice* the appropriate number. The two mechanisms are the `<base>` tag introduced [here](https://github.com/leanprover/verso/blob/nightly-testing/src/verso-html/Main.lean#L307) and I believe [in this calculation](https://github.com/leanprover/verso/blob/nightly-testing/src/verso/Verso/Code/Highlighted.lean#L1338).

jcreedcmu:
Fixed on `nightly-testing` by #556.

PR #556, opened 2025-10-07
fix: make verso-html output correct relative paths

jcreedcmu:
Not sure what the right solution is, but this is an approximation of it.

It seems to me, given that we're already emitting a <base> everywhere that says where the site root is, that we probably want to be emitting relative `href`s relative to that. For the ToC in the sidebar, we already were constructing them as such, but for the body links we're using `Verso.Multi.Path.link`. As a proof of concept, bluntly modifying `Verso.Multi.Path.link` to output relative paths as I did here does fix the symptom, but it's not clear to me that that's the right thing to do.


src/verso-html/Main.lean
```diff
@@ -667,7 +672,7 @@ where
     let mut hash := 0
     let mut n := 0
     while h : n < s.utf8ByteSize do
-      hash := hash + s.getUtf8Byte ⟨n⟩ h
+      hash := hash + s.getUTF8Byte ⟨n⟩ h
```
jcreedcmu: This is just an incidental deprecation fix.

david-christiansen:
This is something that deserves a Verso-wide think-through, I think.

I think it makes a lot of sense to do as follows, for **all genres**, and document this as an expectation for any user-made genres:
1. All links generated by Verso are relative to an ambient site root, rather than absolute or relative to the current page. Absolute hrefs in the current site anywhere but in the `<base>` tag are bugs. Pages that want to link to their parents must track their path rather than using `../`.
2. The site root is specified once, in `<base>`, and the JS code always uses this and fail if it's not present. We get rid of the `__versoSiteRoot` field.
3. Nested pages specify `<base>` relatively.

Does that make sense to you? Right now, it's a bit of a hodge-podge and it gets unclear. This would be a simple rule to explain, though a bit annoying to get implemented.

One downside of this approach is that trailing slash canonicalization becomes something we need to be opinionated about. That may be an issue - not all hosts make that easy.

david-christiansen:
@robsimmons Your thoughts are very welcome here as well

jcreedcmu:
> This is something that deserves a Verso-wide think-through, I think.
> 
> I think it makes a lot of sense to do as follows, for **all genres**, and document this as an expectation for any user-made genres:
> 
>     1. All links generated by Verso are relative to an ambient site root, rather than absolute or relative to the current page. Absolute hrefs in the current site anywhere but in the `<base>` tag are bugs. Pages that want to link to their parents must track their path rather than using `../`.
> 
>     2. The site root is specified once, in `<base>`, and the JS code always uses this and fail if it's not present. We get rid of the `__versoSiteRoot` field.
> 
>     3. Nested pages specify `<base>` relatively.

Yes, good, I like this solution a lot.

> One downside of this approach is that trailing slash canonicalization becomes something we need to be opinionated about. That may be an issue - not all hosts make that easy.

It's not clear to me yet how the issue of whether paths are expressed as relative or absolute affects trailing slash canonicalization.

Here's what I would have thought we'd want to say about trailing slashes, picking "yes trailing slashes" for example as the opinionated stance:

- links we generate to things that could be construed as logically directories are generated as `href="foo/bar/baz/"` and not `href="foo/bar/baz"` or `href="foo/bar/baz/index.html"`
- we include a `<link rel="canonical" href="foo/bar/baz/">` in the file `foo/bar/baz/index.html` that we output
- we refrain from doing emitting any client-side redirect code
- if the author chooses to host their html somewhere that supports configuring server-side redirects of "foo/bar/baz" to "foo/bar/baz/", that's fine, if they don't that's also fine.

By "fine" I mean, I believe that this arrangement will have the consequence that search engine indexing will not have multiple copies of the same page, even if external websites out of our control happen to link to noncanonical URLs on the site. Am I missing any other negative consequences of this arrangement?

jcreedcmu:
David in discussion out of band answered my question of what the interaction between `<base>` and slashes is: it's that if `<base>`'s href itself is relative, (which we want it to be, at least in the common case) then whether the current url of the page does or doesn't have a trailing slash is important for interpreting that relative path.

So current plan for this PR is now:
- Ensure that all html pages generated, across all genres, have a `<base>` which expresses where the site root is relative to this page, as a relative path.
- Ensure all `<a href="...">` and all `fetch` calls are naming paths relative to the siteRoot, with trailing slashes when appropriate.
- Do client-side redirects in javascript from slashless to slash URLs when appropriate (this is what we already are doing)
- Use `<link rel="canonical" href="foo/bar/baz/">` with a trailing slash so that crawlers that don't run javascript still give SEO juice to the appropriate page.

jcreedcmu:
As follow-up work, add an option to specify the site root as an absolute path.

jcreedcmu:
I'd like to review just the more wide-reaching change here to how links are handled in verso, and defer generating `<link rel="canonical" href="...">` to another PR since I expect that to be easy and rather independent.

I manually tested by doing the following:

Manual genre:
- `lake exe usersguide --output inner/_out`
- `python3 -m http.server --directory inner`
- Browse to http://localhost:8000/_out/html-multi/
- Confirm that sidebar links work, content links work, hovers work, and search works as expected.

Blog genre:
- `lake exe demosite --output inner/_site`
- `python3 -m http.server --directory inner`
- Browse to http://localhost:8000/_site/
- Confirm that all links work, hovers work, no unexpected 404s in console.

Literate html:
- `lake build VersoBlog:literate`
- `lake exe verso-html .lake/build/literate inner/literate`
- `python3 -m http.server --directory inner`
- Browse to http://localhost:8000/literate/VersoBlog/
- Confirm that sidebar links work, content links work, hovers work, and search works as expected.

jcreedcmu:
Note that this PR should resolve #555 when landed.

robsimmons:
My fix in #578 is implicitly a strong endorsement @david-christiansen 's proposed standard-relative-links plan above

src/multi-verso/MultiVerso/Path.lean
```diff
@@ -24,10 +24,15 @@ namespace Path
...
-  "/" ++ String.join (path.toList.map (· ++ "/")) ++
+public def relativeLink (path : Path) (htmlId : Option String := none) : String :=
+  String.join (path.toList.map (· ++ "/")) ++
   (htmlId.map ("#" ++ ·)).getD ""
 
+/--
+Retrieves a string that can be used as an absolute link.
+-/
+public def link (path : Path) (htmlId : Option String := none) : String :=
+  "/" ++ relativeLink path htmlId
```
david-christiansen, line 36: I think we can and should do without this operator. It seems like a footgun - we want documents to be hostable under arbitrary path prefixes.
```suggestion
```

src/multi-verso/MultiVerso/Path.lean
```diff
@@ -24,10 +24,15 @@ namespace Path
 /--
 Retrieves a string that can be used as a link.
```
david-christiansen, line 25:
```suggestion
Retrieves a string that can be used as a link.

This is a path that's relative to the site root.
```

src/verso-manual/VersoManual/Html.lean
```diff
@@ -536,7 +536,6 @@ def page
         <title>{{textTitle}}</title>
         <link rel="stylesheet" href="book.css" />
         <link rel="stylesheet" href="verso-vars.css" />
-        <script>s!"const __versoSiteRoot = document.baseURI;"</script>
```
david-christiansen, line 539: 🎉

src/multi-verso/MultiVerso/Path.lean
jcreedcmu: Agree, and since there are a few uses of it I need to spend a little more time understanding, deferring it to https://github.com/leanprover/verso/issues/582.

PR #557, opened 2025-10-07
fix: avoid construction of redundant forward-slashes in search results

### LINKED ISSUE #555: Verso-HTML output has wrong relative paths

robsimmons:
Uses the new URL constructor to create addresses; avoids the redundant-forward-slash problem. Closes #555

### DIFF
--- a/static-web/find.js
+++ b/static-web/find.js
@@ -1,9 +1,9 @@
 let params = new URLSearchParams(document.location.search);
 let domains = params.getAll("domain");
-let name = params.get("name");
+let paramName = params.get("name");
 console.log("Domains: " + domains);
-console.log("name: " + name);
-if(name) {
+console.log("name: " + paramName);
+if(paramName) {
     let siteRoot = typeof __versoSiteRoot !== 'undefined' ? __versoSiteRoot : "";
     let options = [];
     if (domains && domains.length > 0) {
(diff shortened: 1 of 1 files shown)


static-web/find.js
```diff
@@ -1,9 +1,9 @@
 let params = new URLSearchParams(document.location.search);
 let domains = params.getAll("domain");
-let name = params.get("name");
+let paramName = params.get("name");
```
robsimmons, line 3: The variable renaming here is incidental; JavaScript gives a deprecation warning if you use `name` as an identifier outside of a function scope.

static-web/find.js
```diff
@@ -42,12 +42,13 @@ if(name) {
             document.querySelector("#message").innerHTML = "<p>Searched domains:</p>" + "<ul>" + domains.map(x => "<li><code>" + x + "</code>: " + xref[x]['title'] + "</li>\n").join('') + "</ul>";
         });
     } else if (options.length == 1) {
-        let addr = siteRoot + options[0]['address'] + "#" + options[0]['id'];
+        const addr = new URL(options[0]['address'], siteRoot);
+        addr.hash = options[0]['id'];
         window.location.replace(addr);
```
robsimmons, line 47: This is the key change.

Issue #569, opened 2025-10-13
Allow links to https://live.lean-lang.org/ from code blocks


jcreedcmu:
A document author might like to be able to add some metadata to code blocks so that they produce links to https://live.lean-lang.org/ .

Ideally an implementation would work for multiple genres.

More specific instances of this pattern might be:

- The author writes a fully hermetic #mwe code block that includes all necessary imports, and which is fully elaborated/checked/produces-red-squigglies-on-error, but does not affect (nor is affected by!) the surrounding environment

- The author writes a sequence of code blocks (interspersed with narrative text) that together form one hermetic unit. The author's intent that these form one thread could be expressed by some sort of `(name := foobar)` argument to the code block. Each such block could render with a link pointing to an instance of https://live.lean-lang.org/ with all the code up to and including that block. Each code block in this thread could have `-show` set, in which case it does not appear in the rendered HTML document, but would appear in the sandbox code.

- The author writes a branching tree of code blocks (interspersed with narrative text). It's less clear how the author would express the tree they have in mind. This might be easier if we require a "thin" tree that only has one main branch, and digressions from it that only last one block. In that case we could have the reuse the `keep` to determine whether the main branch is added to.

jcreedcmu:
@david-christiansen suggested, instead of using names to connect together different blocks, using an enclosing directive to group them together, e.g. something like

````
:::example

If you do
```lean
def foo := 0
```
and then
```lean
def bar := foo
```
it works 
:::
````

would lead to two lean codeblocks which would know they were part of the same running example. I think I like this approach. One might imagine live-lean links attached to both blocks, each one of which link to a live-lean with `#codez` anchor containing the prefix of the running example up to that block.

Technically, the enclosing directive approach is less general, since if you explicitly named the previous example your block was meant to append, you could interleave two distinct running examples, but I think even as a reader I would find that sort of narrative presentation more confusing than helpful, and am so quite comfortable not supporting it.

jcreedcmu:
It's also the case that a tree of diverging branches of a running example could be handled by nesting of directives.

PR #570, opened 2025-10-13
feat: allow links to live.lean-lang.org

jcreedcmu:
Add a `+liveLink` boolean option to code blocks in the manual genre which causes links to https://live.lean-lang.org/ to appear. 

This is progress towards https://github.com/leanprover/verso/issues/569

### DIFF
--- a/src/verso/Verso/Code/LzCompress.lean
+++ b/src/verso/Verso/Code/LzCompress.lean
@@ -1,3 +1,8 @@
+/-
+Copyright (c) 2025 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Jason Reed
+-/
 import Std.Data.HashMap
 
 def keyStrBase64 := "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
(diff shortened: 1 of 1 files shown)


david-christiansen:
I think the utility of a facility for linking only a single block to live is pretty limited. In all the real documents I know of, it's a _sequence_ of blocks that makes up an example, like:
````
Given this definition:
```lean
def x := 5
```
this proof succeeds:
```lean
theorem x_is_five : x = 5 := by rfl
```
````
I worry that this feature ends up being a footgun, where users write:
````
Given this definition:
```lean +liveLink
def x := 5
```
this proof succeeds:
```lean +liveLink
theorem x_is_five : x = 5 := by rfl
```
````
and the second link gives broken code. It needs some way of tracking its dependencies, I think, like "part of the same example" or "mention block names explicitly".

robsimmons:
I think I generally agree with @david-christiansen 's assessment. It wouldn't be ridiculous to have background state that is tracking the accumulated lean code in a document like we've done on various versions of the Twelf Wiki, but we also presumably need a way to append code to the top of the file too.

jcreedcmu:
I 100% agree threaded examples are desirable, and are discussed on https://github.com/leanprover/verso/issues/569

jcreedcmu:
Thanks both for comments, I've moved my attempt to https://github.com/leanprover/reference-manual/pull/609

PR #578, opened 2025-10-17
fix: relativize paths before navigating to them

robsimmons:
I'd misremembered/interpreted/understood the [URL Web API](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and the [<base> tag spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/base) — if you give an absolute path (starting with a forward slash) then all these will navigate relative to the website root — the full base URL in both cases only matters for relative paths.

Because VersoManual's find page **is** setting the website base to be the "root-of-the-manual", we can just avoid constructing the full URL and navigate to the appropriate relative path, but first we have to strip the slash (that we probably shouldn't be storing in the first place) from the search result.

Also adjusts the leanfile to somewhat-adequately capture the actual text-file dependencies of Verso.

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -28,7 +28,9 @@ input_file mathJs where
 input_dir vendorPopper where
   path := "vendored-js/popper"
 
--- Needs recursive input_dir capability
+-- Commented out because nested directories aren't handled correctly
+-- FIXME: https://github.com/leanprover/lean4/issues/10827
+-- (A single vendoredJs target is preferable when that issue is resolved)
 -- input_dir vendorTippy where
 --   path := "vendored-js/tippy"
 -- input_dir vendorKatex where


robsimmons:
I've temporarily based the v4.24.0 tag of reference-manual off this branch and all seems to be working well!

robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/73bbf1113811758212d36b57bf073c241e609554?parent=5a9318d373ee1c5dfc7443952b30f913ec36298b) for 73bbf1113811758212d36b57bf073c241e609554 against 5a9318d373ee1c5dfc7443952b30f913ec36298b are in! @robsimmons

robsimmons:
(just a for-fun benchmark)

lakefile.lean
```diff
@@ -17,11 +17,32 @@ input_file versoVars where
...
+  path := "static-web/find.js"
+
+input_file mathJs where
+  text := true
+  path := "static-web/math.js"
+
+input_dir vendorPopper where
+  path := "vendored-js/popper"
+
+-- Needs recursive input_dir capability
```
jcreedcmu: Recommend filing issue against Lake, assigning Mac, and putting something like `FIXME(#ISSUE_NUMBER)` here

robsimmons: Done!

robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/602b371768608ff004914505267c9b8aa8c1d5a4?parent=5a9318d373ee1c5dfc7443952b30f913ec36298b) for 602b371768608ff004914505267c9b8aa8c1d5a4 against 5a9318d373ee1c5dfc7443952b30f913ec36298b are in! @robsimmons

PR #580, opened 2025-10-20
refactor: streamline and add comments to addBlock

robsimmons:


src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -450,56 +450,66 @@ def PartElabM.currentLevel : PartElabM Nat := do return (← getThe State).curre
...
-    if ty.isAppOf ``HasLink || ty.isAppOf ``HasNote then pure #[(e, ty)] else pure #[]
-  | .lam _ t b _ | .forallE _ t b _ => do return (← findLinksAndNotes t) ++ (← findLinksAndNotes b)
-  | .letE _ t d b _ => do return (← findLinksAndNotes t) ++ (← findLinksAndNotes d) ++ (← findLinksAndNotes b)
-  | .mdata _ e | .proj _ _ e => findLinksAndNotes e
+    if ty.isAppOf ``HasLink || ty.isAppOf ``HasNote then
+      pure #[(e, ty)]
+    else
+      -- Question RJS: do we actually want to just fall through here?
+      -- Do we every want to leave an mvar?
+      pure #[]
```
robsimmons: @david-christiansen is there a use case for allowing other mvars in the expressions for addblock?

david-christiansen: The error might be an as-yet-unsynthesized type class instance, to be solved by the later call to `synthesizeSyntheticMVarsNoPostponing`.

robsimmons: How could such a type class instance arise in the Verso context?

david-christiansen: Verso extensions can generate arbitrary Lean code that computes blocks or inlines. That might involve something like a `Functor` or `OfNat` instance.

PR #594, opened 2025-10-26
refactor: create an intermediate 'VersoDoc' data structure

robsimmons:
A `VersoDoc genre` is intended to represent an *unevaluated* `Part genre`. The result of a `#doc` command stops being syntax denoting a value of type `Part genre`, and becomes syntax denoting a value of type `VersoDoc genre`.

Presenting this indirection layer allows us for more flexibility in serializing some or all of a document in a way that minimally affects existing code. There are two alternative designs here that I considered:

1) Having `VersoDoc.toPart` be monadic (returning `m (Part genre)` instead of `Part genre`), since we might in the future want to have a part that includes a reference to an auxiliary file or data structure. Doing this seemed like it would require much more substantial changes to existing code.
2) Having `VersoDoc.toPart` require an extra argument, a piece of context that can be threaded through that might represent something like serialization data in the future. This would be somewhat simpler to implement (it would be implementing the reader monad instead of an arbitrary monad).

### DIFF
--- a/src/verso/Verso/Doc/Elab/Monad.lean
+++ b/src/verso/Verso/Doc/Elab/Monad.lean
@@ -218,8 +218,8 @@ partial def FinishedPart.toTOC : FinishedPart → TOC
   | .included name => .included name
 
 /--
-Creates a term denoting a `VersoDoc` value from a `FinishedPart`. This is the final step in turning
-a parsed verso doc into syntax.
+Creates a term denoting a {lean}`VersoDoc` value from a {lean}`FinishedPart`. This is the final step
+in turning a parsed verso doc into syntax.
 -/
 def FinishedPart.toVersoDoc [Monad m] [MonadQuotation m]
     (genreSyntax : Term)


src/verso-blog/VersoBlog/LiterateModuleDocs.lean
```diff
@@ -88,8 +88,8 @@ def elabFromModuleDocs (x : Ident) (path : StrLit) (mod : Ident) (title : StrLit
       }
     pure name
 
-  let metadata ← if let some m := metadata? then `(some $m) else `(none)
-  elabCommand <| ← `(command|def $x : Part $genre := {$(mkIdent mod) with metadata := $metadata})
+  -- let metadata ← if let some m := metadata? then `(some $m) else `(none)
+  elabCommand <| ← `(command|def $x : VersoDoc $genre := $(mkIdent mod))
```
robsimmons: This needs to be fixed for this feature to be merged: right now it's ignoring the metadata because the way the syntax-versus-expr juggling was happening requires some kind of modification here.

robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/c2a86155de3b2274f6c30aab49ef97609890f96d?parent=c15ef5d13de1f1fed2b0b9487acf0889084a9855) for c2a86155de3b2274f6c30aab49ef97609890f96d against c15ef5d13de1f1fed2b0b9487acf0889084a9855 are in! @robsimmons
<details open>
<summary>Major changes (3)</summary>

- `refman-no-opt-args/compile//success` changed.
- `refman-o0/compile//success` changed.
- `refman/compile//success` changed.
</details>

Garmelon:
By the way, the results of this benchmark run may not be comparable because it's using the bundled clang.

Edit: I think the parent commit has already been re-run, so you can disregard the warning.

src/verso/Verso/Doc.lean
```diff
@@ -624,6 +624,40 @@ private partial def Part.reprPrec [Repr genre.Inline] [Repr genre.Block] [Repr g
 public instance [Repr g.Inline] [Repr g.Block] [Repr g.PartMetadata] : Repr (Part g) where
   reprPrec := private Part.reprPrec
 
+/--
+The result type of values created by Verso's `#doc` and `#docs` commands. A value of type
+{lean}`VersoDoc` represents a not-fully-evaluated document of type {lean}`Part` that can be turned
+into a value by invoking the `VersoDoc.toPart` method.
+-/
+public structure VersoDoc (genre : Genre) where
+  private construct : Unit → Part genre
```
david-christiansen: Why is this not a `Thunk (Part genre)`? I expect we'll only evaluate it once in most cases anyway, but I can imagine situations where it might get evaluated multiple times (e.g. generating multiple output formats from the same input).

david-christiansen: Also, why do we need a private constructor and explicit interface here?

src/verso/Verso/Doc.lean
```diff
@@ -624,6 +624,40 @@ private partial def Part.reprPrec [Repr genre.Inline] [Repr genre.Block] [Repr g
...
+
+instance : Inhabited (VersoDoc genre) where
+  default := VersoDoc.mk fun () => Inhabited.default
+
+/--
+Constructs a new VersoDoc; this function wraps the private {lean}`VersoDoc.mk` constructor. Only
+code generated by {lit}`Concrete.lean ` should refer to this function.
```
david-christiansen:
```suggestion
code generated by {lit}`Concrete.lean` should refer to this function.
```

src/verso/Verso/Doc.lean
```diff
@@ -624,6 +624,40 @@ private partial def Part.reprPrec [Repr genre.Inline] [Repr genre.Block] [Repr g
...
+A {lean}`VersoDoc` represents a potentially-not-fully-evaluated {lean}`Part`. Calling {lean}`VersoDoc.toPart` forces
+evaluation of the {lean}`VersoDoc` to a {lean}`Part`.
+-/
+public def VersoDoc.toPart: VersoDoc genre → Part genre
+  | .mk construct => construct ()
+
+/--
+Replace the metadata in a VersoDoc.
+
+This is something of a hack used as a workaround in LiterateModuleDocs.
```
david-christiansen, line 656: This seems like something we should be able to dispense with... What about moving the metadata line higher up in that file?

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -197,11 +194,12 @@ private def linkRefName [Monad m] [MonadQuotation m] (docName : Name) (ref : TSy
 private def footnoteRefName [Monad m] [MonadQuotation m] (genre : Term) (docName : Name) (ref : TSyntax `str) : m Term :=
   ``(HasNote.contents $(quote ref.getString) $(quote docName) (genre := $genre) (self := _))
 
-
-open Lean.Parser.Term in
-partial def FinishedPart.toSyntax [Monad m] [MonadQuotation m] [MonadEnv m]
+/--
+From a finished part, construct a Part value
```
```suggestion
From a finished part, constructs syntax that denotes its `Part` value.
```

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -212,52 +210,24 @@ partial def FinishedPart.toSyntax [Monad m] [MonadQuotation m] [MonadEnv m]
...
-  StateT.run' (ReaderT.run (helper finishedPart) rootName) {}
+  | .included name => ``(VersoDoc.toPart $name)
 
 partial def FinishedPart.toTOC : FinishedPart → TOC
   | .mk titleStx _titleInlines titleString _metadata _blocks subParts endPos =>
     .mk titleString titleStx endPos (subParts.map toTOC)
   | .included name => .included name
 
+/--
+Create a term denoting a VersoDoc value from a FinishedPart
```
```suggestion
Creates a term denoting a `VersoDoc` value from a `FinishedPart`.
```

david-christiansen: And perhaps a sentence or two on when/where/why?

src/verso/Verso/Doc/Concrete.lean
```diff
@@ -212,8 +212,9 @@ private def finishDoc (genre : Term) (title : StrLit) : Command.CommandElabM Uni
   let finished := partElabState.partContext.toPartFrame.close endPos
 
   let n := mkIdentFrom title (← currentDocName)
-  let docu ← finished.toSyntax genre
-  Command.elabCommand (← `(def $n : Part $genre := $docu))
+  let docu ← finished.toVersoDoc genre
```
david-christiansen: Most of the code abbreviates this `doc`. I think this is fine, but we should be consistent when we can to make it easier to puzzle out code.

src/verso/Verso/Doc.lean
robsimmons: Why not thunk: because we're going to migrate this to something more complex.

Why private constructor: probably I'm just being too cute. BUT, I really want the implementation of VersoDoc to be a part of our private API and we are allowed to change it.

robsimmons: Removed the private constructor, possibly valuable but overcomplicated for what we're doing right now.

PR #600, opened 2025-10-29
feat: tutorials genre

david-christiansen:
Adds a genre for tutorials.

Tutorials are very much like the manual genre, and re-use most of its infrastructure and extensions. The differences are:
 1. The rendering of tutorials is more flexible, designed to fit in to existing web designs (and in particular, the Lean design).
 2. Tutorials have different metadata, and will not initially support non-HTML output.

Additionally, tutorials will feature support for loading their content on live.lean-lang.org and downloading their example code.


david-christiansen:
TODO:
 * [x] Test with content in Lean manual repository
 * [x] Live links
 * [x] Tested project tarballs
 * [x] Mutual cross-references with manual-genre documents in same repository

PR #605, opened 2025-10-31
refactor: thread a "document reconstruction" structure

robsimmons:
This structure will eventually hold the redundancy elimination structure, but this PR is just trying to explain how that structure can get threaded through all of the separately-addAndCompile-d blocks

### DIFF
--- a/src/verso/Verso/Doc.lean
+++ b/src/verso/Verso/Doc.lean
@@ -659,7 +659,7 @@ Replace the metadata in a VersoDoc.
 This is something of a hack used as a workaround in LiterateModuleDocs.
 -/
 public def VersoDoc.withMetadata (metadata? : Option genre.PartMetadata)  : VersoDoc genre → VersoDoc genre
-  | .mk construct => .mk fun recon => { construct recon with metadata := metadata? }
+  | .mk construct => .mk fun docReconst => { construct docReconst with metadata := metadata? }
 
 /--
 Identify function; this is a temporary compatibility shim to introduce a new type,
--- a/src/verso/Verso/Doc/Elab/Monad.lean
+++ b/src/verso/Verso/Doc/Elab/Monad.lean
@@ -141,7 +141,7 @@ deriving Inhabited
 
 def DocElabContext.fromGenreTerm (genreSyntax : Term) : TermElabM DocElabContext := do
   let genre ← Term.elabTerm genreSyntax (some (.const ``Doc.Genre []))
-  return DocElabContext.mk genreSyntax genre .always (.some <| mkIdent (← mkFreshUserName `docRecon))
+  return DocElabContext.mk genreSyntax genre .always (.some <| mkIdent (← mkFreshUserName `docReconst))
 
 structure DocElabM.State where
   linkRefs : HashMap String DocUses := {}
--- a/src/tests/Tests/Elab.lean
+++ b/src/tests/Tests/Elab.lean
@@ -42,7 +42,7 @@ def totallyUndefined : RoleExpanderOf Unit
 /--
 error: don't know how to synthesize placeholder for argument `head`
 context:
-docReconInBlock✝ : Doc.DocReconstruction
+docReconstInBlock✝ : Doc.DocReconstruction
 ⊢ Doc.Inline Doc.Genre.none
 -/
 #guard_msgs in
--- a/src/verso/Verso/Doc/Elab.lean
+++ b/src/verso/Verso/Doc/Elab.lean
@@ -291,7 +291,7 @@ where
       let msg := m!"Block content found in a context where a header was expected."
       let note := MessageData.note m!"A document part (section/chapter/etc) consists of a header, followed by zero or more blocks, followed by zero or more sub-parts. This block occurs after a sub-part{which}, but outside of the sub-parts."
       throwErrorAt cmd "{msg}\n{note}"
-    let hygenicName := some <| mkIdent (← mkFreshUserName `docReconInBlock)
+    let hygenicName := some <| mkIdent (← mkFreshUserName `docReconstInBlock)
     let blk ← withTheReader DocElabContext ({ · with docReconstructionPlaceholder := hygenicName }) <|
       elabBlock cmd
     addBlock blk (blockInternalDocReconstructionPlaceholder := hygenicName)


robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/765b99e23c544d3814d1eb4dc41570d2a087b6bd?parent=dc5ba59a9eccd111ea649f526c10d0b3ff4ef892) for 765b99e23c544d3814d1eb4dc41570d2a087b6bd against dc5ba59a9eccd111ea649f526c10d0b3ff4ef892 are in! @robsimmons
<details>
<summary>Minor changes (3)</summary>

- `refman-no-opt-args/build/.total//generated C` changed by **-1.4%** (✅).
- `refman-o0/build/.total//generated C` changed by **-1.4%** (✅).
- `refman/build/.total//generated C` changed by **-1.4%** (✅).
</details>

src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -170,7 +172,7 @@ quoting a compressed version of the highlighted code.
 -/
 private def quoteHighlightViaSerialization (hls : Highlighted) : DocElabM Term := do
   let repr := hlToExport hls
-  ``(hlFromExport! $(quote repr))
+  ``(reconstructHighlight (_ : DocReconstruction) $(quote repr))
```
david-christiansen: Why is an underscore suitable here? I'd expect to be getting "can't  synthesize placeholder" errors, so this confuses me when reading.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -127,11 +127,13 @@ structure DocElabContext where
   genre : Expr
 
   refsAllowed : RefsAllowed
+
+  docReconstructionRef : Option Ident
```
```suggestion
  /-- Whether references to not-yet-defined footnotes and links are permitted -/
  refsAllowed : RefsAllowed
  /-- The name of an `IORef` that holds mutable state during elaboration which is later accessible in the entire document. -/
  docReconstructionRef : Option Ident
```
Or something like that - this last docstring expresses my confusion more than any confidence about what this does.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -416,6 +423,22 @@ unsafe def inlineExpandersForUnsafe (x : Name) : DocElabM (Array InlineExpander)
 @[implemented_by inlineExpandersForUnsafe]
 opaque inlineExpandersFor (x : Name) : DocElabM (Array InlineExpander)
 
+def substForDocReconstructionMvars (same : Expr) (e : Expr) : MetaM Expr := do match e with
```
david-christiansen: Oh, is this why the `_` is OK? Subtle! Perhaps there should be some dedicated syntax that just macro-expands to the underscore but has a docstring expressing what's going on?

david-christiansen: Why is the parameter named `same`? What's the intuition there?

src/verso/Verso/Doc.lean
```diff
@@ -631,32 +631,35 @@ private partial def Part.reprPrec [Repr genre.Inline] [Repr genre.Block] [Repr g
...
+
 /--
 The result type of values created by Verso's `#doc` and `#docs` commands. A value of type
 {lean}`VersoDoc` represents a not-fully-evaluated document of type {lean}`Part` that can be turned
 into a value by invoking the `VersoDoc.toPart` method. The actual structure of a {lean}`VersoDoc`
 should not be relied on.
 -/
 public structure VersoDoc (genre : Genre) where
-  construct : Unit → Part genre
+  construct : DocReconstruction → Part genre
```
david-christiansen, line 643: Presumably this will be:
```suggestion
  construct : genre.DocReconstruction → Part genre
```
in the end, right? And then `DocElabM` will have one of these in its state?

robsimmons: Hmm... I was thinking of making DocReconstruction initially special-cased (just to redundancy elimination) and then dynamic (which is what we'd want to do elaboration-time, as in your Syntax-to-Eval PR), but you're right that it could be a genre type component, that hadn't occurred to me!

I think I'm filing that under the case for "special-case to redundancy elimination for right now because we don't know what the extensible API really ought to be"

src/verso/Verso/Doc/Elab/Monad.lean
robsimmons: Once upon a time I was using MetaM to force the mvars to be equal, so all the mvars were getting identified with the same thing, hence "same." But this is unenlightening; `replacement` is better than `same`.

src/verso/Verso/Doc/Elab/Monad.lean
robsimmons: Added comments in 54e9f24ca

robsimmons: Changed the nature of the argument, and correspondingly changed explanation, in [28e09bd](https://github.com/leanprover/verso/pull/605/commits/28e09bd554548c8022375205b442799a4faba39d)

src/verso-manual/VersoManual/InlineLean.lean
robsimmons: I've changed how this works to refer to a specific unbound identifier instead of this weird mvar trick.

src/verso/Verso/Doc/Elab/Monad.lean
robsimmons: I think I finally sorted out how to do something concrete here, so this is irrelevant now

robsimmons:
@david-christiansen I modified the structure of the PR quite a bit after sleeping on it and thinking about your comments — the current structure doesn't use MVars at all anymore and so consequently doesn't need to traverse the Expr. I'll merge next week.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -126,12 +126,22 @@ structure DocElabContext where
...
   genre : Expr
 
+
+  /-- Whether references to undefined (not-yet-defined) footnotes and links are permitted. -/
   refsAllowed : RefsAllowed
+
+  /--
+  The docReconstructionPlaceholder provides a free variable during Verso document elaboration. This
+  syntax object cannot be successfully elaborated to a term until closed as a function
+  {lit}`` `(fun $docReconstructionPlaceholder => $termContainingFreeVariable)``.
```
```suggestion
  {lit}`` `(fun $docReconstructionPlaceholder => $termContainingFreeVariable) ``.
```
Verso inherits this from Markdown: if there's both a leading and trailing space in a code element, then it is removed.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -126,12 +126,22 @@ structure DocElabContext where
...
+  -/
+  docReconstructionPlaceholder : Option Ident
 deriving Inhabited
 
+
 def DocElabContext.fromGenreTerm (genreSyntax : Term) : TermElabM DocElabContext := do
   let genre ← Term.elabTerm genreSyntax (some (.const ``Doc.Genre []))
-  return DocElabContext.mk genreSyntax genre .always
+  return DocElabContext.mk genreSyntax genre .always (.some <| mkIdent (← mkFreshUserName `docRecon))
```
david-christiansen: I wrote this on the other one, but I figure it makes sense here too.

In my head, "recon" bascially always abbreviates "reconnaissance" rather than "reconstruction", which lead to me always needing to do a double-take on this abbreviation.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -143,6 +153,7 @@ structure PartElabM.State where
   linkDefs : HashMap String (DocDef String) := {}
   footnoteDefs : HashMap String (DocDef (Array (TSyntax `term))) := {}
   deferredBlocks : Array (Name × Term) := #[]
+
```
```suggestion
```
We don't typically have a newline here

david-christiansen:
I think this implementation is much easier to follow.

src/verso/Verso/Doc/Elab/Monad.lean
robsimmons: Sorry to well-actually this, but Verso isn't actually following Markdown (at least the CommonMark standard) here — which trims at most one space at the front or at the back. It's not a requirement that there be spaces on both sides. I'd actually prefer if Verso changed to follow commonmark here, but for now this is a good change.

src/verso/Verso/Doc/Elab/Monad.lean
robsimmons: changing to "reconst" as an abbreviation

src/verso/Verso/Doc/Elab/Monad.lean
david-christiansen: This isn't how I interpret CommonMark. See e.g. [example 332](https://spec.commonmark.org/0.31.2/#example-332). What am I getting wrong?

PR #613, opened 2025-11-03
docs: better readme for using the literate facet

robsimmons:


README.md
```diff
@@ -279,23 +279,45 @@ Lean code to HTML. This HTML includes hovers, clickable "go to
...
-1. Generate literate program data from your Lean libraries or modules
+1. Add Verso as a package dependency.
+
+   If you're using a `lakefile.toml` configuration, that looks like this:
+
+   ```toml
+   [[require]]
+   name = "verso"
+   git = "https://github.com/leanprover/verso.git"
+   rev = "latest"
```
david-christiansen, line 292: We should encourage users to pick a tag that corresponds to their Lean version here, esp. because this is oriented towards users for whom Verso/docs are not the main use case.

README.md
```diff
@@ -279,23 +279,45 @@ Lean code to HTML. This HTML includes hovers, clickable "go to
...
+   ```
+
+   If you're using a `lakefile.lean` configuration, that looks like this:
+
+   ```lean4
+   require verso from git "https://github.com/leanprover/verso.git"@"latest"
```
david-christiansen, line 298: Likewise here - it should use a tag that matches the toolchain.

README.md
```diff
@@ -279,23 +279,45 @@ Lean code to HTML. This HTML includes hovers, clickable "go to
...
    ```
    lake build MyLib:literate
    ```
 
-2. Generate HTML from this literate program data. To put the HTML in
-   `htmldir`, run:
+   This generates files in the folder `.lake/build/literate`
+
+3. Generate HTML from this literate program data. To put the HTML in a
+   directory named `dist`, run:
```
david-christiansen: This seems like a weird directory name to me. Though only slightly weirder than `htmldir`. What about `html`?

README.md
```diff
@@ -279,23 +279,45 @@ Lean code to HTML. This HTML includes hovers, clickable "go to
...
    
    ```
-   lake exe verso-html ./lake/build/literate htmldir
+   lake exe verso-html .lake/build/literate dist
    ``` 
 
+You can preview the resulting files by running
+`python3 -m http.server -d dist` and pointing a web browser to
```
david-christiansen: I think these command lines should always include an explicit port number, just so users who need to change it (and will probably know what it is) won't have to look up the argument syntax for this Python module, which is a bit hard to find.

README.md
robsimmons: Is there a way to signal this that *doesn't* feel like adding an extra concept that the user has to understand or an effective two additional steps that the user has to perform (step one: "open your lean-toolchain file and note the 'tag', the part after the colon. If your lean-toolchain file contains `leanprover/lean4:v4.25.0-rc2`, then the tag is `v4.25.0-rc2`, step two: "replace the word `"REVISION"` in the sample above with your lean tag, so `"v4.25.0-rc2"` in the example we gave".)

README.md
robsimmons: `dist` is the current popular kid at the lunch table for putting build artifacts that are supposed to be distributed, especially web artifacts (it's the Vite default). But `html` is quite sensible.

README.md
david-christiansen: This is a thing that they have to understand all over the place. It would be nice to make that unnecessary, but I don't think this repo is the place to do it.

david-christiansen: (and the thing after the : isn't really the tag _per se_, it's the Lean release. We just use tags that correspond to Lean versions)

README.md
david-christiansen: In my head, `dist` is where the source code goes for a source tarball that's going to constitute a release.

README.md
robsimmons: Fair enough; added a paragraph suggesting replacing latest with their Lean version.

README.md
robsimmons: fair but definitely not what the Vite kids are doing. Changed to `html` in any case!

README.md
robsimmons: added, now I have a place to look up this information myself (the readme that I wrote 😄 )

PR #614, opened 2025-11-03
fix: reset hearbeat count for each block

robsimmons:
At the suggestion of @Kha , avoid the apparent need for bumping heartbeat counts that I had encountered and fixed in leanprover/reference-manual#644 by just resetting the heartbeat count upon each block installation task. Reverts the reference manual benchmarking version to a pre-heartbeat-bumping variant

### DIFF
--- a/.reference_manual_revision
+++ b/.reference_manual_revision
@@ -1,3 +1,3 @@
 # This should be a version of https://github.com/leanprover/reference-manual
 # Used to benchmark the performance of verso on the reference manual
-2bfa3475e5e83b8e087622d43d93166bd071e55c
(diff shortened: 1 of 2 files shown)


robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/9e25922592751d90c7dfcd4b03bb5d1d431c10bf?parent=5b16ba298987d0780e0bd8a24c8a449112ac265e) for 9e25922592751d90c7dfcd4b03bb5d1d431c10bf against 5b16ba298987d0780e0bd8a24c8a449112ac265e are in! @robsimmons

PR #626, opened 2025-11-14
feat: implement basic syntax highlighting

jcreedcmu:
For now, have a separate function for converting `Highlighted` into TeX that can rely on it being used in the context of a `\Verbatim` environment, meaning that it doesn't need to wrap every token in, e.g., something like `\texttt`. Possibly we may want to move everything to using this in-verbatim-context function, but I haven't committed to it yet.

Use standard TeX logger

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,7 @@ jobs:
             environ
             etoolbox
             ulem
+            newunicodechar
 
       - name: Check `tlmgr` version
         run: tlmgr --version
--- a/.github/workflows/merge-main-nightly.yml
+++ b/.github/workflows/merge-main-nightly.yml
@@ -103,6 +103,8 @@ jobs:
             pgf
             environ
             etoolbox
+            ulem
+            newunicodechar
 
       - name: Check `tlmgr` version
         run: tlmgr --version
--- a/.github/workflows/update-nightly.yml
+++ b/.github/workflows/update-nightly.yml
@@ -160,7 +160,8 @@ jobs:
           pgf
           environ
           etoolbox
-
+          ulem
+          newunicodechar
 
     - name: Check `tlmgr` version
       run: tlmgr --version


src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+| c, .sort _ => .raw c
+| c, .levelVar _ => .raw c
+| c, .levelConst _ => .raw c
+| c, .moduleName _ => .raw c
+| c, .levelOp _ => .raw c
+| c, .withType _ => .raw c
+| c, .unknown => .raw c
+
+/--
+Replace characters with strings simultaneously
```
david-christiansen:
```suggestion
Replaces characters with strings simultaneously.
```

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+      have : (String.Pos.Raw.next s pos).byteIdx > pos.byteIdx :=
+        String.Pos.Raw.byteIdx_lt_byteIdx_next s pos
+      let c := pos.get s
+      let s' := match replace c with | some rs => rs | none => s!"{c}"
+      loop (acc ++ s') (pos.next s)
+    termination_by s.endPos.1 - pos.1
+  loop "" 0
+
+/--
+Escape a string in an appropriate way for uses of
```
david-christiansen:
```suggestion
Escapes a string in an appropriate way for uses of
```

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+  -- is that DejaVu Sans Mono Oblique lacks the former, and we'd like
+  -- to render variables in italic/oblique. If we decide on a
+  -- different monospace font that has better coverage, we could
+  -- reconsider this decision.
+  -- Ordinarily it'd be pretty sketchy to subvert author intent by replacing
+  -- characters that users might otherwise copy and paste into Lean,
+  -- but this is exactly the character Lean inserts to create shadowed
+  -- names, so it seems benign to change it in this case.
+  | '✝' => some "†"
+  | _ => none)
```
david-christiansen:
```suggestion
  replaceChars s fun
  | '{' => some "\\symbol{123}"
  | '|' => some "\\symbol{124}"
  | '}' => some "\\symbol{125}"
  | '\\' => some "\\symbol{92}"
  -- The reason for changing U+271d LATIN CROSS to U+2020 DAGGER here
  -- is that DejaVu Sans Mono Oblique lacks the former, and we'd like
  -- to render variables in italic/oblique. If we decide on a
  -- different monospace font that has better coverage, we could
  -- reconsider this decision.
  -- Ordinarily it'd be pretty sketchy to subvert author intent by replacing
  -- characters that users might otherwise copy and paste into Lean,
  -- but this is exactly the character Lean inserts to create shadowed
  -- names, so it seems benign to change it in this case.
  | '✝' => some "†"
  | _ => none
```

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+  -- The reason for changing U+271d LATIN CROSS to U+2020 DAGGER here
+  -- is that DejaVu Sans Mono Oblique lacks the former, and we'd like
+  -- to render variables in italic/oblique. If we decide on a
+  -- different monospace font that has better coverage, we could
+  -- reconsider this decision.
+  -- Ordinarily it'd be pretty sketchy to subvert author intent by replacing
+  -- characters that users might otherwise copy and paste into Lean,
+  -- but this is exactly the character Lean inserts to create shadowed
+  -- names, so it seems benign to change it in this case.
+  | '✝' => some "†"
```
david-christiansen: These look quite different in some fonts. Can't we configure TeX to use a different fallback font for just this character?

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+/-- info: "\\symbol{123}\\symbol{124}\\symbol{125}\\symbol{92}" -/
+#guard_msgs in
+#eval escapeForVerbatim "{|}\\"
+
+defmethod Highlighting.Token.toVerbatimTeX (t : Highlighting.Token) : Verso.Output.TeX :=
+  highlightToken (escapeForVerbatim t.content) t.kind
+
+open Verso.Output.TeX in
+/--
+Return TeX that is appropriate for the content of a \Verb environment (from package fancyvrb)
```
david-christiansen:
```suggestion
Returns TeX that is appropriate for the content of a `\Verb` environment (from package `fancyvrb`)
```

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+      let c := pos.get s
+      let s' := match replace c with | some rs => rs | none => s!"{c}"
+      loop (acc ++ s') (pos.next s)
+    termination_by s.endPos.1 - pos.1
+  loop "" 0
+
+/--
+Escape a string in an appropriate way for uses of
+`\Verb[commandchars=\\\{\}]|...|` and
+`\begin{Verbatim}[commandchars=\\\{\}]...\end{Verbatim}`
```
david-christiansen: Below, you phrase this as "with command characters ..." which seems a bit easier to read. Why is this phrased differently?

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -12,14 +12,164 @@ open Verso.Output
...
+  | .token t => highlightToken (escapeForVerbatim t.content) t.kind
+  | .text str => .raw (escapeForVerbatim str)
+  | .seq hts => .seq <| hts.map (·.toVerbatimTeX)
+  | .span info content =>
+     if h : info.size > 0
+     then .seq #[.raw s!"\\{info[0].1.toString}Decorate\{", content.toVerbatimTeX, .raw "}"]
+     else content.toVerbatimTeX
+  | .tactics _info _start _end content => content.toVerbatimTeX
+  | .point _kind _info => \TeX{"[Point]"}
+  | .unparsed str => .raw (escapeForVerbatim str) -- FIXME: Is this right?
```
david-christiansen: I think it's right - why wouldn't it be? This shows up when highlighting text that includes a parser error, and we can't highlight things that don't parse but we do want to include them.

src/verso/Verso/Code/HighlightedToTex.lean
jcreedcmu: I looked into it before the PR, but couldn't find an easy way. Is <a href="https://tex.stackexchange.com/questions/514940/define-fallback-font-for-missing-glyphs-in-lualatex">this approach more or less what you had in mind?</a> This seems like more lua programming than I have an appetite for maintaining. Is there a simpler way I'm missing?

Separately, I believe "latin cross might look different from dagger in font X", but also "latin cross in font Y might look quite different from the other characters in font X", so it's not completely clear to me what font substitution is buying us if we don't have control over X and Y. As it is, I am specifying Deja Vu Sans Mono. If there were an easy way on the TeX side to substitute Deja Vu Sans Mono (not Oblique) latin cross, which does exist, for Deja Vu Sans Mono Oblique latin cross, I'd be happy with that.

src/verso/Verso/Code/HighlightedToTex.lean
jcreedcmu: No, good catch, this was a change I forgot to apply everywhere.

src/verso/Verso/Code/HighlightedToTex.lean
jcreedcmu: Thanks for explaining where `.unparsed` arose in practice! That was the information I was missing.

src/verso/Verso/Code/HighlightedToTex.lean
david-christiansen: I had remembered incorrectly - in my mind, `fontspec` had a way to associate different Unicode ranges with different fonts.

It does look like `\newunicodechar{✝}{{\OTHERFONT ✝}}` may do the trick here for this character.

I wonder if we should do some experiments and see if we can configure Iosevka's metrics to match the main fonts we want to use? A custom Iosevka could get us slab serifs (to make code easier to distinguish from non-code while still feeling reasonably current design-wise), good math Unicode, and decent-looking text that mentions identifiers, all at a fairly low amount of work. I'll look into that when I get a minute.

jcreedcmu:
I wouldn't mind Iosevka.
I can confirm in https://typeof.net/Iosevka/specimen that it has U+271D. I was worried for a second that my local copy didn't have it, but maybe they added it since I last downloaded/generated it.

I'm going to try generating a couple of slab variants at sensible weights/italicness and see how it looks. Vendoring these the same way that we're vendoring KaTeX fonts makes sense, yeah?

jcreedcmu:
Hm, perhaps slightly different things need to happen after the decision to vendor, but I do think vendoring the generated Iosevka fonts makes more sense than expanding the dependencies we're exposing to our users to include everything needed to build iosevka.

jcreedcmu:
Hmm at least by default Iosevka doesn't have ✝ match the widths of the rest of the characters of the monospaced font. Trying to see if there's an easy fix.

jcreedcmu:
Hm, no easy solution that I can see at the moment. I'm still inclined towards punting on the Right Solution to U+271D LATIN CROSS and going with my hacky U+2020 DAGGER substitution for now and revisiting later.

PR #649, opened 2025-12-03
fix: bitrot in verso-html

david-christiansen:
Also adds a test to prevent it in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Generate some source HTML
         run: |
-          lake build Verso.Hover:litearte
+          lake build Verso.Hover:literate
           lake exe verso-html .lake/build/literate htmlout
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -75,6 +75,11 @@ jobs:
         run: |
           lake exe simplepage
 
+      - name: Generate some source HTML
+        run: |
+          lake build Verso.Hover:litearte
```
jcreedcmu:
```suggestion
          lake build Verso.Hover:literate
```

.github/workflows/ci.yml
```diff
@@ -75,6 +75,11 @@ jobs:
...
+          lake exe verso-html .lake/build/literate htmlout
```
jcreedcmu, line 81:
```suggestion
          lake exe verso-html .lake/build/literate htmlout
          test -f htmlout/Verso/Hover/index.html
```

PR #654, opened 2025-12-10
fix: compile on macOS Tahoe 26.1

lancelet:
The version of Lean being used in some of the examples was suffering from issue #5649 "macOS clang fails to link against dylib generated by Lean": https://github.com/leanprover/lean4/issues/5649

This PR bumps the versions so that they compile correctly on macOS.

(I'm investigating this project for the first time. I'm not sure if my proposed changes here harm `subverso` usage / testing, but `lake build` fails for me in the `main` branch without them.)

### DIFF
--- a/examples/website-examples/Examples.lean
+++ b/examples/website-examples/Examples.lean
@@ -44,7 +44,7 @@ def foo (n k : Nat) : Nat :=
   if n < k then
     1 + foo (n + 1) k
   else 0
-termination_by foo n k => k - n
+termination_by k - n
 %end
 
 %example version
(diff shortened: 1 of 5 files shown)


examples/website-examples/Examples.lean
```diff
@@ -44,7 +44,7 @@ def foo (n k : Nat) : Nat :=
   if n < k then
     1 + foo (n + 1) k
   else 0
-termination_by foo n k => k - n
+termination_by k - n
```
lancelet, line 47: I'm not sure that this is old syntax anymore... I had to change it to compile with the version of Lean I bumped.

david-christiansen:
SubVerso has its own tests, so it's fine to update these. Thanks!

PR #657, opened 2025-12-10
feat: give search suggestions based on `@[suggest_for]` annotations

robsimmons:
Closes leanprover/reference-manual#588

### DIFF
--- a/src/verso-manual/VersoManual/Docstring.lean
+++ b/src/verso-manual/VersoManual/Docstring.lean
@@ -809,8 +809,8 @@ def docstring.descr : BlockDescr := withHighlighting {
       }}
 
   localContentItem := fun _id info _contents => open Verso.Output.Html in do
-    let  (name, _declType, _signature, _customLabel) ←
-      FromJson.fromJson? (α := Name × Block.Docstring.DeclType × Signature × Option String) info
+    let  (name, _declType, _signature, _customLabel, _altNames) ←
+      FromJson.fromJson? (α := Name × Block.Docstring.DeclType × Signature × Option String × Array Name) info
     let names := #[name.getString!, name.toString]
     pure <| names.map fun s => (s, {{<code>{{s}}</code>}})
 


src/verso-manual/VersoManual/Docstring.lean
```diff
@@ -1349,13 +1389,15 @@ def docstring : BlockCommandOf DocstringConfig
...
 
-
       let declType ← Block.Docstring.DeclType.ofName name (hideFields := hideFields) (hideStructureConstructor := hideCtor)
 
       let signature ← Signature.forName name
 
       let extras ← getExtras name declType
-      ``(Verso.Doc.Block.other (Verso.Genre.Manual.Block.docstring $(quote name) $(quote declType) $(quote signature) $(quote customLabel)) #[$(blockStx ++ extras),*])
+
+      let altNames ← getStoredSuggestions name
```
robsimmons, line 1404: This is the line that won't work until the toolchain is bumped.

src/verso-manual/VersoManual/Docstring.lean
```diff
@@ -687,17 +687,49 @@ open Verso.Search in
...
+}.setFont { family := .code }
+
 open Verso.Genre.Manual.Markdown in
 @[block_extension Block.docstring]
 def docstring.descr : BlockDescr := withHighlighting {
   init st := st
     |>.setDomainTitle docstringDomain "Lean constant reference"
     |>.setDomainDescription docstringDomain "Documentation for Lean constants"
     |>.addQuickJumpMapper docstringDomain docDomainMapper
+    |>.setDomainTitle suggestionDomain "Lean constant suggestion"
```
david-christiansen:
```suggestion
    |>.setDomainTitle suggestionDomain "Lean constant suggestions"
```

src/verso-search/VersoSearch/DomainSearch.lean
```diff
@@ -54,6 +54,16 @@ public structure DomainMapper where
   CSS to be used in the quick-jump box to customize results from the given domain.
   -/
   quickJumpCss : Option String := none
+  /--
+  JavaScript code to give a custom rendering for a searchable item. This should be a function that
+  takes three arguments:
+
+  {open DomainMapper}
+   * {lit}`searchable` is a Searchable object (returned by the {name}`dataToSearchables` function)
+   * {lit}`matchedParts` is an array of objects with TypeScript type {lit}`{t: 'text', v: string} | {t: 'highlight', v: string}`
```
david-christiansen: I think it's useful to provide a bit more semantics here:
```suggestion
   * {lit}`matchedParts` is an array of objects with TypeScript type {lit}`{t: 'text', v: string} | {t: 'highlight', v: string}`
     Objects with `text` in the `t` field are the parts of the search result that provide context, while objects with `highlight` are the parts of the result that match the search string.
```

src/verso-search/VersoSearch/DomainSearch.lean
```diff
@@ -54,6 +54,16 @@ public structure DomainMapper where
   CSS to be used in the quick-jump box to customize results from the given domain.
   -/
   quickJumpCss : Option String := none
+  /--
+  JavaScript code to give a custom rendering for a searchable item. This should be a function that
+  takes three arguments:
+
+  {open DomainMapper}
+   * {lit}`searchable` is a Searchable object (returned by the {name}`dataToSearchables` function)
```
david-christiansen, line 62:
```suggestion
   * {lit}`searchable` is a `Searchable` object (returned by the {name}`dataToSearchables` function). `Searchable` is the TypeScript type {lit}`{searchKey: string, address: string, domainId: string, ref?: any}`. `searchKey` is the user's  search text, `address` is the result's link destination, `domainId` is the name of the Verso domain, and `ref` is present for full-text results and contains the internal document reference for the full-text index. 
```

david-christiansen: Though this makes me wonder if this function is ever called for full-text results. It probably shouldn't be, in which case perhaps `Searchable` should be restricted to not include `ref`. Maybe the custom rendering function should be curried and just take `searchKey`, `address`, and `domainId` as parameters?

robsimmons: I believe this is the correct description:

> `ref` is arbitrary data and can be used to pass information from `dataToSearchables` to the `customRender` function

There is a weirdness where ref is being populated by other `dataToSearchables` functions even though it is, I am relatively confident, being unused. But this PR is critically using `ref` to smuggle data from searchable creation to the custom render function.

src/verso-manual/VersoManual/Docstring.lean
david-christiansen: This has happened now. Should we update and try for a merge?

robsimmons: yep!

PR #659, opened 2025-12-10
chore: require Prettier formatting for html/js/css and markdown

robsimmons:
Uses Prettier to format anything Prettier is willing to format (CSS, HTML, JS, and Markdown). Set up to allow for automatic triggering on push-to-main, and by-hand triggering if desired, which minimizes the need for anyone to have Prettier actually installed on their machine.

I tried to match existing practice — 4-space tabs are more common than 2-space tabs in the codebase, and double-quoted strings match Lean more than single-quoted strings. Markdown files are formatted (mostly this just does word-wrapping) at 70 columns, and everything else at 100 column.s

An example of the result of the code action:
https://github.com/leanprover/verso/commit/a73e050bc66e807e6d733e53c9681e34d2bd434a?w=1#diff-c574d97f49ae51f97a4a47c2c853880cbeaf83764a6d10e7d7aaf572c739cd9d

### DIFF
--- a/static-web/search/search-box.js
+++ b/static-web/search/search-box.js
@@ -27,8 +27,10 @@ const searchIndex = /** @type {{searchIndex: TextSearchIndex}} */ (/** @type {un
 /**
  * @type {Record<string, DocContentPromise>}
  */
-const docContents =
-    /** @type {any} */ (window).docContents || /** @type {any} */ (window.docContents = {});
+const docContents = ((/** @type {any} */ windowAny) => {
+    if (!windowAny.docContents) windowAny.docContents = {};
+    return windowAny.docContents;
+})(window);
 
 /** Whether to search word prefixes or whole words in full-text searches. Should match the setting in search-highlight.js.
  * @type {boolean}


david-christiansen:
Do I understand right that this auto-formats and commits after each push?

I'm a bit afraid that this will make `git blame` harder to use. I think there's two good solutions, though:
1. Set up the formatting infrastructure to add format-only commits to an `ignoreRevs` file. This would require two commits per format run, though. We should do this for an initial commit that runs prettier on everything, at least.
2. Require that things be formatted the way the tool wants, rather than having the tool do it and commit. Maybe using something like [reviewdog suggestions](https://github.com/reviewdog/reviewdog#code-suggestions) on PRs?

I think number 2 is less likely to cause confusing history issues in the long run, and it's also not so onerous to ask contributors to either install prettier or manually implement what CI tells them is missing. It's not a hard thing to set up, and as you point out, there's good integration everywhere. As long as users don't need it, we're good.

david-christiansen:
We could also have CI require matching formatting, and provide a bot that formats a PR when users write `!format` or somesuch.

ejgallego:
> We could also have CI require matching formatting, and provide a bot that formats a PR when users write `!format` or somesuch.

I'd prefer this solution, and have `lake format` or the editors to take care of the formatting on the developer side.

I've had good experiences with auto-formatters in the past, and something I like it that they usually make review easier; having an extra commit adds a bit of noise IMO for example when doing review using the Github interface.

robsimmons:
Are we ok with requiring formatting for now, and creating the bot after the fact? @david-christiansen @ejgallego

robsimmons:
For what it's worth "just require PRs get the formatting right" is my usual approach anyway, I was trying to avoid even the appearance of adding a dependency. So I definitely don't feel like my arm is being twisted to switch up approaches to checking instead of creating commits 😄

There is an option in the github action to *rewrite the last commit* to format... but that's spooky I don't like it (and doesn't seem necessary in context).

david-christiansen:
Maybe the prettier check should be a separate workflow, leaving `ci.yml` untouched? That'd get us better parallelism when there's enough runners, and faster failures when it's violated.

robsimmons:
@david-christiansen separate makes sense to me

PR #666, opened 2025-12-11
feat: more tex output support for fpil

jcreedcmu:
Add more TeX/PDF output functionality specifically with an eye towards supporting Functional Programming in Lean. In more detail,
- Add some Reader data to the `TeXT` monad so that we can detect whether we're in a fragile environment, and change our output appropriately. This fixed some problems with characters that needed to escaped in section headers (e.g. "Running IO Actions With #eval") where we can't get away with verbatim directly.
- Move several TeX output helpers from pure to TeXT-monadic to make this possible.
- Implement TeX output for the "sample file" block in Manual, which FPiL uses.
- Refactor some helpers to handle newlines better.

### DIFF
--- a/src/tests/Tests/TexUtil.lean
+++ b/src/tests/Tests/TexUtil.lean
@@ -38,4 +38,4 @@ def toTex (block : Doc.Block Genre.Manual) : IO Output.TeX := do
   }
 
   -- Convert the block to TeX
-  block.toTeX.run ⟨options, traverseContext, traverseState, default⟩ |>.run extension_impls
+  block.toTeX.run ⟨options, traverseContext, traverseState, {}⟩ |>.run extension_impls


src/verso-manual/VersoManual.lean
```diff
@@ -287,8 +287,9 @@ def emitTeX (logError : String → IO Unit) (config : Config) (text : Part Manua
   let authors := text.metadata.map (·.authors) |>.getD []
   let date := text.metadata.bind (·.date) |>.getD ""
   let ctxt := {logError}
-  let frontMatter ← text.content.mapM (·.toTeX (opts, ctxt, state))
-  let chapters ← text.subParts.mapM (·.toTeX (opts, ctxt, state))
+  let tex_ctxt := default
```
david-christiansen: As a matter of style, I'd normally use `{}` here for "all fields have the default value", and use `default` for cases when I really want to exercise the `Inhabited` instance, like `panic!`.

david-christiansen: We also don't typically `snake_case` for local variables:
```suggestion
  let texCtxt := default
```

src/verso-manual/VersoManual/Docstring.lean
```diff
@@ -541,8 +541,11 @@ def internalSignature.descr : BlockDescr where
     open Verso.Output.TeX in do
     let .ok (name, signature) := FromJson.fromJson? (α := Highlighted × Option Highlighted) info
       | Verso.Doc.TeX.logError "Failed to deserialize docstring section data while generating TeX"; return .empty
-    let signatureTeX := if let .some sig := signature then \TeX{ " : " \Lean{sig.toTeX}} else .empty
-    pure \TeX{\par " " \Lean{name.toTeX} \Lean{signatureTeX} \Lean{.seq (← contents.mapM goB)}}
+    let signatureTeX ← do
+      if let some sig := signature
+      then pure \TeX{ " : " \Lean{ (← sig.toTeX) } }
+      else pure .empty
```
david-christiansen: Usual Lean formatting is:
```suggestion
      if let some sig := signature then
        pure \TeX{ " : " \Lean{ (← sig.toTeX) } }
      else pure .empty
```

src/verso-manual/VersoManual/Docstring.lean
```diff
@@ -605,10 +608,11 @@ def fieldSignature.descr : BlockDescr where
       | .public => .empty
       | .private => \TeX{ \textbf{"private"} }
       | .protected => .empty
-    let desc := \TeX{ \par " " \Lean{visibility} \Lean{name.toTeX} " : " \Lean{signature.toTeX} \par " " \Lean{.seq (← contents.mapM goB)}}
+    let desc := \TeX{ \par " " \Lean{visibility} \Lean{← name.toTeX} " : " \Lean{← signature.toTeX} \par " " \Lean{.seq (← contents.mapM goB)}}
+    let parentsTeX := (← parents |>.toList |>.mapM (·.toTeX)).intersperse (.raw ", ")
```
david-christiansen:
```suggestion
    let parentsTeX := (← parents.toList.mapM (·.toTeX)).intersperse (.raw ", ")
```

src/verso-manual/VersoManual/InlineLean/IO.lean
```diff
@@ -194,7 +194,34 @@ block_extension Block.exampleLeanFile (filename : String) where
...
   traverse _ _ _ := pure none
-  toTeX := none
+
+  toTeX :=
+    open Verso.Output.TeX in
+    open Verso.Doc.TeX in
+    some <| fun _ _ _ data blocks => do
+      match FromJson.fromJson? (α := FileType) data with
+      | .error err =>
+        IO.println <| "Couldn't deserialize file metadata while rendering HTML: " ++ err
```
david-christiansen:
```suggestion
        IO.println <| "Couldn't deserialize file metadata while rendering TeX: " ++ err
```

david-christiansen: And maybe the TeX and HTML renderers should share the part of the code that deserializes the `data` argument, and just differ in their output?

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -118,45 +114,59 @@ public defmethod Highlighted.toVerbatimTeX : Highlighted → Verso.Output.TeX
...
-  | .trace _cls _msg _children _collapsed => .empty -- FIXME: what to render here?
-  | .append mcs => .seq (mcs.map messageContentsToVerbatimTeX)
 
-public defmethod Highlighted.Message.toTeX (h : Highlighted.Message) : Verso.Output.TeX :=
+  let toRow (h : Highlighted.Hypothesis Highlighted) : TeXT g m Verso.Output.TeX  := do
+    let namesTeX := (← h.names.mapM (·.toTeX)).toList.intersperse (.text " ")
+    pure <| .seq #[namesTeX, .text " : ", (← h.typeAndVal.toTeX)]
+  rows := rows ++ (← hypotheses.mapM toRow)
+  rows := rows ++ #[verbatim (goalPrefix) ++ (← conclusion.toTeX)]
+  pure \TeX{\begin{tabular}{"l"} \Lean{rows.map (· ++ .raw "\\\\")} \end{tabular}}
```
david-christiansen:
```suggestion
  pure \TeX{\begin{tabular}{"l"} \Lean{rows.map (· ++ .raw r#"\\"#)} \end{tabular}}
```

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -118,45 +114,59 @@ public defmethod Highlighted.toVerbatimTeX : Highlighted → Verso.Output.TeX
...
+
+def messageContentsToVerbatimTeX [Monad m] [GenreTeX g m] (h : Highlighted.MessageContents Highlighted) : TeXT g m Verso.Output.TeX :=
+  match h with
+  | .text str => pure str
+  | .goal g => pure g.toString
+  | .term t => pure t.toVerbatimTeX
+  | .trace _cls _msg _children _collapsed => pure .empty -- FIXME: what to render here?
```
david-christiansen, line 156: For TeX, the results should show the children if `collapsed` is true. I think the goal should be to match how it looks in the InfoView reasonably well. Maybe some kind of nested tabular or nested custom list-like environment?

It should probably always show `[cls] msg`, and then if `collapsed = false`, show the children underneath, nested and indented.

src/verso/Verso/Code/HighlightedToTex.lean
```diff
@@ -118,45 +114,59 @@ public defmethod Highlighted.toVerbatimTeX : Highlighted → Verso.Output.TeX
...
+  | .append mcs => do
+      -- We are doing this two-step dance only because lean can't see termination
+      -- when we directly call mcs.mapM messageContentsToVerbatimTeX. This probably
+      -- wouldn't be necessary if there were appropriate @[wf_preprocess] lemmas for mapM.
```
david-christiansen, line 160: There's no need to be afraid of `partial` for this kind of code. Termination is nice, but it's not really essential when we're not proving anything else about it.

src/verso/Verso/Doc/TeX.lean
```diff
@@ -26,7 +34,7 @@ public def Options.lift [MonadLiftT m m'] (opts : Options g m) : Options g m' :=
   opts.reinterpret MonadLiftT.monadLift
 
 public abbrev TeXT (genre : Genre) (m : Type → Type) : Type → Type :=
-  ReaderT (Options genre m × genre.TraverseContext × genre.TraverseState) m
+  ReaderT (Options genre m × genre.TraverseContext × genre.TraverseState × TeXContext) m
```
david-christiansen, line 36: Why not nest `ReaderT`here? Then you can use `readThe` to get individual fields.

jcreedcmu: Fair point, I preferred to not have a mix of products and ReaderT for data that's still pertinent to TeXT, and the accessor already exists.

src/verso/Verso/Code/HighlightedToTex.lean
jcreedcmu: Fair point, but I did enjoy learning about wf_preprocess as a side effect :smile:

src/verso/Verso/Code/HighlightedToTex.lean
jcreedcmu: Done.

src/verso-manual/VersoManual/Docstring.lean
jcreedcmu: Done.

src/verso-manual/VersoManual/Docstring.lean
jcreedcmu: Done.

src/verso-manual/VersoManual.lean
jcreedcmu: Done and done.

PR #677, opened 2025-12-17
feat: add unparsed data to lean codeblocks

robsimmons:
This PR steals a trick from the error explanations infrastructure that's necessary to reproduce [this error explanation](https://lean-lang.org/doc/reference/latest/Error-Explanations/lean___inductionWithNoAlts/#The-Lean-Language-Reference--Error-Explanations--lean___inductionWithNoAlts) with the existing lean-code-fence elaborator. 

When I reuse the existing `lean` code fence elaborator for error explanations without this PR, this error example, with a parse error, renders like this:

```lean
theorem zero_mul (m : [Nat](http://localhost:8880/Basic-Types/Natural-Numbers/#Nat___zero)) : 0 * m = 0 := by
  [induction](http://localhost:8880/Tactic-Proofs/Tactic-Reference/#induction) m with n n_ih
  rw [Nat.mul_zero]
  rw [Nat.mul_succ]
  rw [Nat.add_zero]
  rw [n_ih]
```

renders like this:
image

with this change, it renders like this:
image

### DIFF
--- a/src/verso-manual/VersoManual/InlineLean.lean
+++ b/src/verso-manual/VersoManual/InlineLean.lean
@@ -198,7 +198,9 @@ private def toHighlightedLeanBlock (shouldShow : Bool) (hls : Highlighted) (str:
 
   let range := Syntax.getRange? str
   let range := range.map (← getFileMap).utf8RangeToLspRange
-  ``(Block.other (Block.lean $(← quoteHighlightViaSerialization hls) (some $(quote (← getFileName))) $(quote range)) #[Block.code $(quote str.getString)])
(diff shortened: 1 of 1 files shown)


robsimmons:
> Should there be a systematic check of all calls to `highlight`? 

@jrr6 was describing a reason why he didn't think we should just replace highlight with highlightIncludingUnparsed that he discussed in leanprover/subverso#138 — but I'm not sure I understand the reason right now. I've put a post it note on my wall to remind me to look at this again and investigate this highlighting code further in the new year.

PR #679, opened 2025-12-18
fix: detect and cancel a race condition in search

robsimmons:
[~discussion on FRO zulip~](https://lean-fro.zulipchat.com/#narrow/channel/532109-verso/topic/Search.20concurrency.20bug/near/564500169)[#verso > Search concurrency bug @ 💬](https://lean-fro.zulipchat.com/#narrow/channel/532109-verso/topic/Search.20concurrency.20bug/near/564500169) 

Here is said discussion:

In Playwright testing, I observed the following bug state after programmatically typing "Htmlnone":

image

there's two sets of search results, the 1/1 we expect at the top followed by a bunch of other search results

this then looks like your standard clear-twice-then-write-twice kind of bug. Medium-high confidence:

- you clear [here](https://github.com/leanprover/verso/blob/eb15d7ddc082ce4dcf0a6bfbba560d857600f5c9/static-web/search/search-box.js#L682)
- you release your lock [here](https://github.com/leanprover/verso/blob/eb15d7ddc082ce4dcf0a6bfbba560d857600f5c9/static-web/search/search-box.js#L790) with an await
- you append child [here](https://github.com/leanprover/verso/blob/eb15d7ddc082ce4dcf0a6bfbba560d857600f5c9/static-web/search/search-box.js#L778), [here](https://github.com/leanprover/verso/blob/eb15d7ddc082ce4dcf0a6bfbba560d857600f5c9/static-web/search/search-box.js#L802), and [here](https://github.com/leanprover/verso/blob/eb15d7ddc082ce4dcf0a6bfbba560d857600f5c9/static-web/search/search-box.js#L819)

This actually helps me understand why the bug always appears at the same place when it occurred — the happenstance ordering of search results means that "Html." as a search string happened to grab a bunch of extra domains, so [the sum of these loadbucket calls](https://github.com/leanprover/verso/blob/eb15d7ddc082ce4dcf0a6bfbba560d857600f5c9/static-web/search/search-box.js#L315-L339) take many more milliseconds to resolve for the "Html." search string than for the "Html.none" search string.

It's not enough to fix the clear-then-yield-then-write ordering bug, we actually need to make sure we **cancel** search results for expired queries, because when you rapidly type "Html.none" — at least, at Playwright speeds — it's likely that the "Html." search results will resolve **after** the "Html.none" search results

### DIFF
--- a/static-web/search/search-box.js
+++ b/static-web/search/search-box.js
@@ -745,6 +745,7 @@ class SearchBox {
         allResults.sort((x, y) => y.score - x.score);
         allResults = allResults.slice(0, 30);
 
+        // NOTE: Cancellable speculative execution segment begins here! (No awaits before this.)
         // The following computation is async, and so might fall behind later search results if
         // the user types quickly. This counter detects whether the current function has become
         // stale due to a later-starting search request


.github/workflows/ci.yml
```diff
@@ -181,9 +181,8 @@ jobs:
                   key: playwright-${{ runner.os }}-${{ hashFiles('browser-tests/uv.lock') }}
 
             - name: Install Playwright browsers
-              run:
-                  uv run --project browser-tests --extra test playwright install chromium firefox
-                  --with-deps
+              run: |
+                  uv run --project browser-tests --extra test playwright install chromium firefox --with-deps
```
robsimmons, line 185: (Unrelated autoformatter-generated issue fix)

browser-tests/test_search.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TestSearch:
+    def test_searchbar(self, server: str, page: Page):
+        """Test that the search box can find suggestions."""
+        page.goto(f"{server}") 
+
+        page.get_by_role("searchbox").type("Verso provides genre authors with tools for generating HTML and TeX code via embedded languages")
+        expect(page.get_by_role("searchbox")).to_match_aria_snapshot('- searchbox "Search": Verso provides genre authors with tools for generating HTML and TeX code via embedded languages')
+        expect(page.get_by_label("Results")).to_match_aria_snapshot("""
+          - listbox "Results":
```
jcreedcmu, line 12: my only thought about this is that it seems like a rather big amount of text to match literally (as opposed to like having a more lenient regex match of just some of the text we expect to show up) and seems like it might likely cause us to have to do a little bit of manual "ok just replace the text with what it is now" for very boringly, incidentally non-orthogonal changes... but maybe that's ok. Maybe we want the sensitive test on purpose to let us know when things are changing.

static-web/search/search-box.js
```diff
@@ -726,25 +734,30 @@ class SearchBox {
...
         allResults = allResults.slice(0, 30);
 
-        this.filteredOptions = [];
-        this.firstOption = null;
-        this.lastOption = null;
+        // The following computation is async, and so might fall behind later search results if
+        // the user types quickly. This counter detects whether the current function has become
+        // stale due to a later-starting search request
+        this.requestCounter += 1;
+        const requestCount = this.requestCounter;
```
jcreedcmu, line 753: Add some comment here explaining that this increment is happening specifically *here* for a reason. (i.e. no awaits above, and the asynchronous operation that deserves a sequence number happens right below, as we discussed)

browser-tests/test_search.py
robsimmons: My actual goal here was to put enough text that I got a *unique* result

jcreedcmu: Ah, gotcha

robsimmons: I managed to find a [correct horse battery staple](https://xkcd.com/936/)-esque shorter search string for the current corpus (syntactic ASTs libraries extensions)

ejgallego: A few test systems I've used in the past (like Dune) have a `update-fixtures` operation that will take care of updating large text outputs (like Asts) when the changes are intentional.

We could implement something like that to ease our lives with output texts if we need.

PR #680, opened 2025-12-19
fix: restore intended behavior to suggestion items, add tests

robsimmons:
- change the key for the suggestion domain so that there can be multiple altname -> name mappings
 - actually output the custom renderer code when provided — I know I implemented this before but it somehow got dropped from the previous PR
 - add tests (and some suggestions mostly for the purpose of powering those tests)

### DIFF
--- a/src/verso-manual/VersoManual/InlineLean/IO.lean
+++ b/src/verso-manual/VersoManual/InlineLean/IO.lean
@@ -210,7 +210,7 @@ def Block.exampleFile.descr : BlockDescr := withHighlighting {
           | other =>
             IO.println <| s!"Expected a single code block in an example file, but got {other.size} blocks"
             return .empty
-        let str := str.trimLeft.trimRight
+        let str := str.trimAscii.copy
         let descr : Output.TeX :=
           match type with
           | .stdin => \TeX{\texttt{"stdin"} }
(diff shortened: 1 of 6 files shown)


src/verso-manual/VersoManual/Docstring.lean
```diff
@@ -775,8 +775,8 @@ def docstring.descr : BlockDescr := withHighlighting {
 
     for altName in altNames do
       modify fun st => st
-        |>.saveDomainObject suggestionDomain name.toString id
-        |>.saveDomainObjectData suggestionDomain name.toString (json%{
+        |>.saveDomainObject suggestionDomain (altName.toString ++ "->" ++ name.toString) id
+        |>.saveDomainObjectData suggestionDomain (altName.toString ++ "->" ++ name.toString) (json%{
```
david-christiansen: This is a tiny detail, but perhaps this should be a Unicode arrow with spaces around it? It might make the JSON more readable for anyone implementing a tool.

src/verso/Verso/Output/Html.lean
```diff
@@ -98,6 +98,7 @@ where
 /--
 The empty HTML document.
 -/
+@[suggest_for Verso.Output.Html.nil Verso.Output.Html.none]
```
david-christiansen, line 101: 🎉

david-christiansen (approved):

src/verso-manual/VersoManual/Docstring.lean
robsimmons: done!

PR #691, opened 2025-12-28
EXPERIMENT: rename VersoDoc to DocThunk

robsimmons:
This PR is a refactoring of some of the intermediate data structure created in #594. `VersoDoc` was an extremely generic name, and `DocThunk`, with a corresponding `DocThunk.force : DocThunk genre -> Part genre` function, better communicates what is happening here: computation of a part is getting deferred during the metaprogramming phase until compiled-program-execution time.

The structure of a `DocThunk` is also changed somewhat, to better represent and support the operation of replacing metadata without evaluating the thunk, and to recognize that some applications build `Part` data directly and so don't have a meaningful serialization step.

### DIFF
--- a/src/verso/Verso/Doc/Reconstruct.lean
+++ b/src/verso/Verso/Doc/Reconstruct.lean
@@ -1,0 +1,84 @@
+/-
+Copyright (c) 2023-2024 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Rob Simmons
+-/
+
+module
+public import SubVerso.Highlighting
+public import Verso.Doc
+
+set_option doc.verso true
+set_option pp.rawOnError true
+
+namespace Verso.Doc
+open Lean
+
+public structure DocReconstruction where
+  highlightDeduplication : SubVerso.Highlighting.Export
+
+/--
+The result type of values created by Verso's {lit}`#doc` and {lit}`#docs` commands. A value of type
+{name}`DocThunk` represents a partially-serialized document of type {lean}`Part` that can be turned
+into a value by invoking the `DocThunk.force` method. The internal structure of a {name}`DocThunk`
+should not be relied upon.
+-/
+public inductive DocThunk (genre : Genre) where
+  /--
+  A partially-serialized document. If the {name}`replacementMetadata?` is non-{name}`none`, the
+  value will replace the outermost {name}`Part.metadata` field of the reconstructed document.
+  -/
+  | serialized
+      (construct : DocReconstruction → Part genre)
+      (docReconstructionData : String := "{}")
+      (replacementMetadata? : Option (Option genre.PartMetadata))
+
+  /--
+  For pathways that build {lean}`Part` values directly, there is no value in, or need for, a
+  serialization phase.
+  -/
+  | value (part : Part genre)
+
+instance : Inhabited (DocThunk genre) where
+  default := .value default
+
+/--
+Replace the top-level metadata in a {name}`DocThunk` without forcing the thunk.
+-/
+public def DocThunk.withMetadata (metadata? : Option genre.PartMetadata)  : DocThunk genre → DocThunk genre
+  | .serialized construct docReconstStr _ =>
+    .serialized construct docReconstStr (.some metadata?)
+  | .value part =>
+    .value { part with metadata := metadata? }
+
+/--
+A {name}`DocThunk` represents a potentially-not-fully-evaluated {name}`Part`. Calling
+{name}`DocThunk.force` forces evaluation of the {name}`DocThunk` to a {name}`Part`.
+-/
+public def DocThunk.force: DocThunk genre → Part genre
+  | .serialized construct highlight metadata? =>
+    match Json.parse highlight with
+    | .error e => panic! s!"Failed to parse DocThunk's Export data as JSON: {e}"
+    | .ok json =>
+      let part :=
+        if let .ok highlightJson := json.getObjVal? "highlight" then
+          match SubVerso.Highlighting.Export.fromJson? highlightJson with
+          | .error e => panic! s!"Failed to deserialize Export data from parsed JSON: {e}"
+          | .ok table => construct ⟨table⟩
+        else construct ⟨{}⟩
+      match metadata? with
+      | .none => part
+      | .some replacement => { part with metadata := replacement }
+  | .value part => part
+
+@[deprecated DocThunk (since := "2025-12-28")]
+public def VersoDoc : Genre → Type := DocThunk
+
+@[deprecated DocThunk.force (since := "2025-12-28")]
+public def VersoDoc.force : DocThunk genre → Part genre := DocThunk.force
+
+@[deprecated DocThunk.force (since := "2025-12-28")]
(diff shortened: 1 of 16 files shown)


src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+namespace Verso.Doc
+open Lean
+
+public structure DocReconstruction where
+  highlightDeduplication : SubVerso.Highlighting.Export
+
+/--
+The result type of values created by Verso's {lit}`#doc` and {lit}`#docs` commands. A value of type
+{name}`DocThunk` represents a partially-serialized document of type {lean}`Part` that can be turned
+into a value by invoking the `DocThunk.force` method. The internal structure of a {name}`DocThunk`
```
david-christiansen, line 23: Why not make the constructors private instead of relying on a documentation admonishment?

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+-/
+public inductive DocThunk (genre : Genre) where
+  /--
+  A partially-serialized document. If the {name}`replacementMetadata?` is non-{name}`none`, the
+  value will replace the outermost {name}`Part.metadata` field of the reconstructed document.
+  -/
+  | serialized
+      (construct : DocReconstruction → Part genre)
+      (docReconstructionData : String := "{}")
+      (replacementMetadata? : Option (Option genre.PartMetadata))
```
david-christiansen, line 34: A comment on this field that indicates why it exists would be useful.

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+should not be relied upon.
+-/
+public inductive DocThunk (genre : Genre) where
+  /--
+  A partially-serialized document. If the {name}`replacementMetadata?` is non-{name}`none`, the
+  value will replace the outermost {name}`Part.metadata` field of the reconstructed document.
+  -/
+  | serialized
+      (construct : DocReconstruction → Part genre)
+      (docReconstructionData : String := "{}")
```
david-christiansen, line 33: The docstring for this constructor should say a little bit about what this string is expected to contain.

If we're saving lots of strings with serialized JSON in them, perhaps we can make things easier by defining a wrapper structure? Something like:
```
structure SerializedDocReconstruction where
  json : String

def SerializedDocStructure.empty : SerializedDocReconstruction where
  json := "{}"
```

This should make the type signatures easier to read, and may prevent some confusion.

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+  For pathways that build {lean}`Part` values directly, there is no value in, or need for, a
+  serialization phase.
+  -/
+  | value (part : Part genre)
+
+instance : Inhabited (DocThunk genre) where
+  default := .value default
+
+/--
+Replace the top-level metadata in a {name}`DocThunk` without forcing the thunk.
```
david-christiansen:
```suggestion
Replaces the top-level metadata in a {name}`DocThunk` without forcing the thunk.
```

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+-/
+public def DocThunk.withMetadata (metadata? : Option genre.PartMetadata)  : DocThunk genre → DocThunk genre
```
david-christiansen, line 48: I wonder if this wouldn't be more compositional with a constructor that replaces metadata instead of a field on `serialized`. Do we expect there to be lots of calls that might lead to a measurable overhead?

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+  | .serialized construct docReconstStr _ =>
+    .serialized construct docReconstStr (.some metadata?)
+  | .value part =>
+    .value { part with metadata := metadata? }
+
+/--
+A {name}`DocThunk` represents a potentially-not-fully-evaluated {name}`Part`. Calling
+{name}`DocThunk.force` forces evaluation of the {name}`DocThunk` to a {name}`Part`.
+-/
+public def DocThunk.force: DocThunk genre → Part genre
```
david-christiansen: I am so looking forward to the auto-formatter!
```suggestion
public def DocThunk.force : DocThunk genre → Part genre
```

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+public structure DocReconstruction where
+  highlightDeduplication : SubVerso.Highlighting.Export
+
+/--
+The result type of values created by Verso's {lit}`#doc` and {lit}`#docs` commands. A value of type
+{name}`DocThunk` represents a partially-serialized document of type {lean}`Part` that can be turned
+into a value by invoking the `DocThunk.force` method. The internal structure of a {name}`DocThunk`
+should not be relied upon.
+-/
+public inductive DocThunk (genre : Genre) where
```
david-christiansen, line 26: One way in which this is very much unlike `Thunk` is that repeated calls to `force` will do repeated work. I think this name is better than `VersoDoc`, which says basically nothing, but I wonder whether `DocThunk` is the right name given that we don't cache the result.

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+module
+public import SubVerso.Highlighting
+public import Verso.Doc
+
+set_option doc.verso true
+set_option pp.rawOnError true
+
+namespace Verso.Doc
+open Lean
+
```
david-christiansen, line 16: I think a "why" here as a module docstring would be valuable. Something like:
```suggestion
/-!
In {given}`g : Genre`, the Verso elaborator does not directly emit a {lean}`Part g`. This is because included
Lean code samples tend to be highly repetitive, especially in long proofs with growing contexts. This repetition
can lead to performance problems.

Instead, the elaborator deduplicates the included code samples via hash-consing. The hash-cons table is
saved as a string literal containing a JSON serialization, because large Lean constant data structures generate large C code compared to the string, and the overhead of deserializing it at run-time is negligible compared to the overhead of compiling very large C files (most Verso document code is run only once per compilation).
-/
```

src/verso/Verso/Doc/Reconstruct.lean
```diff
@@ -0,0 +1,84 @@
...
+public structure DocReconstruction where
```
david-christiansen, line 17: A docstring here would be nice:
```suggestion
/--
A deduplication table for code samples embedded in a document, so repeated code (e.g. hypotheses in proof 
states) does not lead to code bloat.
-/
public structure DocReconstruction where
```

Should the constructor of this structure be private?

src/verso/Verso/Doc/Reconstruct.lean
robsimmons: I originally had one but don't actually know how to put a comment on a field of a branch of a datatype. Or should the comment go on `serialized`?

PR #453, opened 2025-06-25
EXPERIMENT: produce Expr directly for Verso docs instead of term syntax

david-christiansen:
This should decrease build times. For now, this branch is a way to test that.

PR #692, opened 2025-12-28
EXPERIMENT: typed syntax as the target of doc elaboration

robsimmons:
This PR allows block elaboration to target a typed concrete representation of blocks instead of producing `Term` objects. It is directly modeled off of #453, which retargeted block elaboration from `Term` to `Expr`. Unlike in that PR, here the goal is not to reduce build times, it is to make document elaboration less metaprogramming-ful, and to move towards being able to serialize much of the structure of an elaborated document as JSON or something similar.

The first critical change is in `Verso.Doc.Elab.Basic`: the `FinishedPart` type, which is already a concrete representation of the result of elaborating a document, is renamed to `Elab.Part`, and corresponding `Elab.Block` and `Elab.Inline` types are added to represent the concrete representations of the result of elaborating a block and of elaborating inline syntax. (`Elab.Inline` is just an alias for `Term`, since we're not retargeting term elaboration in this PR.)

The second critical change is in `Verso.Doc.Elab.Monad`, where a new `block_elab` decorator is added to parallel the `block_expander` decorator, and in `Verso.Doc.Elab.Block`, where the new family of decorators are queried. This is a relatively isolated change: the interesting changes happen when elaborators are introduced that parallel the `code_block_expander` and `directive_expander` decorators.

### DIFF
--- a/src/verso/Verso/Doc/Elab/Basic.lean
+++ b/src/verso/Verso/Doc/Elab/Basic.lean
@@ -163,7 +163,7 @@ parts being built.
 public def PartContext.level (ctxt : PartContext) : Nat := ctxt.parents.size
 
 /--
-Closes the current part. The resulting {name}`Part` is appended to
+Closes the current part. The resulting {name}`FinishedPart` is appended to
 {name (full := PartFrame.priorParts)}`priorParts`, and
 the top of the stack of our parents becomes the current frame. Returns
 {name}`none` if there are no parents.
--- a/src/verso/Verso/Doc/Elab/Basic.lean
+++ b/src/verso/Verso/Doc/Elab/Basic.lean
@@ -20,6 +20,8 @@ public inductive TOC where
   | included (name : Ident)
 deriving Repr, TypeName, Inhabited
 
+/- Elaboration targets -/
+namespace Target
 /--
 The result of elaborating an inline component of a document
 -/
@@ -46,8 +48,8 @@ deriving Repr, BEq
 end
 
 mutual
-public partial def ListItem.toTerm [Monad m] [MonadRef m] [MonadQuotation m] (genre : Term) : Array Block → m Term
-  | contents => do ``(Doc.ListItem.mk #[$(← contents.mapM (Block.toTerm genre)),*])
+public partial def ListItem.toTerm [Monad m] [MonadRef m] [MonadQuotation m] (genre : Term) : Array Target.Block → m Term
+  | contents => do ``(Doc.ListItem.mk #[$(← contents.mapM (Target.Block.toTerm genre)),*])
 
 public partial def DescItem.toTerm [Monad m] [MonadRef m] [MonadQuotation m] (genre : Term) : Array Inline × Array Block → m Term
   | (term, desc) => do ``(Doc.DescItem.mk #[$term,*] #[$(← desc.mapM (Block.toTerm genre)),*])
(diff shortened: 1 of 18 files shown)


robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/f90542ac3071707a76a3591786255c863a21dd23?reference=41cc5c326e5651c0df5618fde11eada549fa6581) for f90542ac3071707a76a3591786255c863a21dd23 against 41cc5c326e5651c0df5618fde11eada549fa6581 are in! @robsimmons
<details open>
<summary>Large changes (9✅)</summary>

- ✅ `refman-no-opt-args/build/.total//c.o time`: **-18m 53s** (**-70.6%**)
- ✅ `refman-no-opt-args/build/.total//eval time`: **-24m 3s** (**-67.5%**)
- ✅ `refman-no-opt-args/build/.total//wall clock time`: **-2m 44s** (**-74.7%**)
- ✅ `refman-o0/build/.total//c.o time`: **-11m 23s** (**-59.7%**)
- ✅ `refman-o0/build/.total//eval time`: **-21m 55s** (**-64.9%**)
- ✅ `refman-o0/build/.total//wall clock time`: **-1m 9s** (**-57.2%**)
- ✅ `refman/build/.total//c.o time`: **-11m 31s** (**-59.8%**)
- ✅ `refman/build/.total//eval time`: **-21m 43s** (**-64.7%**)
- ✅ `refman/build/.total//wall clock time`: **-1m 10s** (**-57.3%**)
</details>

robsimmons:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/verso/commits/052dcb41a5606859df85d4d0cec66ac69f11405b?reference=41cc5c326e5651c0df5618fde11eada549fa6581) for 052dcb41a5606859df85d4d0cec66ac69f11405b against 41cc5c326e5651c0df5618fde11eada549fa6581 are in! @robsimmons

No significant changes detected.

src/verso/Verso/Doc/Elab/Basic.lean
```diff
@@ -20,8 +20,65 @@ public inductive TOC where
   | included (name : Ident)
 deriving Repr, TypeName, Inhabited
 
+/- Elaboration targets -/
+namespace Target
+/--
+The result of elaborating an inline component of a document
+-/
+public abbrev Inline := Term
```
david-christiansen, line 28: Why are these `Term`, rather than a parallel structure to `Target.Block`? Has it just not yet been done, or is there an underlying reason?

src/verso/Verso/Doc/Elab/Basic.lean
```diff
@@ -20,8 +20,65 @@ public inductive TOC where
...
+
+variable (genre : Genre) in
+public section
+/--
+The result of elaborating a document block. Concretely denotes a {lean}`Doc.Block genre` for some
+unspecified {name}`genre`.
+-/
+public inductive Block where
```
david-christiansen, line 36: This looks like it could be implemented without the repetition as:
```
inductive InlineExt where
  | other (container : Term)
  | stx (stx : Term)
inductive BlockExt where
  | other (container : Term)
  | stx (stx : Term)
abbrev Inline := Lean.Doc.Inline InlineExt
abbrev Block := Lean.Doc.Block InlineExt BlockExt
```
That would give the same expressive power, from what I can see, while allowing a lot more shared code.

src/verso/Verso/Doc/Elab/Basic.lean
```diff
@@ -20,8 +20,65 @@ public inductive TOC where
...
+  | dl (items : Array ((Array Inline) × (Array Block)))
+  | blockquote (items : Array Block)
+  | concat (content : Array Block)
+  | other (container : Term) (content : Array Block)
+  | stx (stx : Term)
+deriving Repr, BEq
+
+end
+
+mutual
```
david-christiansen, line 50: This whole mutual block strikes me as code that should be `Quote` instances. That wouldn't let the code use quotations, but this kind of thing isn't so hard to build directly. Is it because there were issues propagating the genre via unification?

src/verso/Verso/Doc/Elab/Block.lean
```diff
@@ -26,28 +28,42 @@ def decorateClosing : TSyntax `block → DocElabM Unit
...
         try
           let termStx ← withFreshMacroScope <| e stx
-          return termStx
+          return .stx termStx
+        catch
+          | ex@(.internal id) =>
+            if id == unsupportedSyntaxExceptionId then continue
+            else throw ex
+          | ex => throw ex
+      let exp ← blockElabsFor kind
```
david-christiansen, line 62: Having this in two stages says that all expanders take precedence over all elaborators. I think the current code follows the usual Lean practice, where more-recently-defined expanders take precedence. I say "I think" because this feature is used in only a few places so it's not fresh in my mind.

I think this is OK, but it should be a deliberate and documented decision rather than an artifact of the implementation. To make them use the same precedence rules, the expander/elab extension could be one extension that contains a sum type instead of two separate ones.

src/verso/Verso/Doc/Elab/Basic.lean
david-christiansen: Oh wait, these are converting from `Target.Block`. The fact that I stumbled here is evidence that at least a comment would be useful!

src/verso/Verso/Doc/Elab/Basic.lean
david-christiansen: Or, alternately,
```
inductive InlineExt where
  | other (container : Term)
  | stx (stx : Term)

inductive BlockExt where
  | other (container : Term)
  | stx (stx : Term)

def Target : Genre where
  Inline := InlineExt
  Block := BlockExt
  PartMetadata := Empty
  TraverseContext := Empty
  TraverseState := Empty
```
but I think that's a bit less straightforward.

david-christiansen:
I think this is a nice direction to go in with a lot of potential to make it much easier to implement Verso extensions.

Given that the Verso AST is "closed" (the concrete syntax is fixed), I think this same approach could be put earlier in the pipeline. Instead of passing `TSyntax block` to the elaborators, what if they got `Source.Block` as parameters? This would be an inductive type of Verso concrete syntax, with all the source positions and such.

src/verso/Verso/Doc/Elab/Basic.lean
robsimmons: Just hasn't been done — was demonstrating a block-only incremental adoption.

PR #693, opened 2025-12-29
EXPERIMENT: elaborate block structures as concrete elements

robsimmons:


src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -799,11 +799,27 @@ public abbrev DirectiveExpander := Array Arg → TSyntaxArray `block → DocElab
 
 public abbrev DirectiveExpanderOf α := α → TSyntaxArray `block → DocElabM Term
 
+public abbrev DirectiveElaborator := Array Arg → TSyntaxArray `block → DocElabM (Array Target.Block)
```
david-christiansen, line 802: Do we want this old-school version to be public?

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -838,14 +854,29 @@ unsafe initialize registerBuiltinAttribute {
 
     let n ← mkFreshUserName <| declName ++ `directive
 
+    -- Find the appropriate adapter for both new (`DirectiveElabOf`) and legacy (`DirectiveExpanderOf`) declarations
+    let constructor ← match (← getEnv).find? declName |>.map (·.type.getAppFn) with
```
david-christiansen, line 858: This is probably correct, but I figured I'd flag it up just to be sure. Right now, this will reject user-written `abbrev`s that expand to one of the allowed forms. The check could be done modulo defeq to allow those. I think this is correct anyway because I think the attributes reject types that don't strictly have this form, but I thought it was worth double-checking.

david-christiansen, line 858: `constructor` has a specific meaning that I think we should avoid overloading. What about `adapter`?

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -865,10 +896,10 @@ unsafe initialize registerBuiltinAttribute {
       directiveExpanderExt.addEntry env (directiveName, n)
 }
 
-private unsafe def directiveExpandersForUnsafe' (x : Name) : DocElabM (Array (DirectiveExpander × Option String × Option SigDoc)) := do
+private unsafe def directiveExpandersForUnsafe' (x : Name) : DocElabM (Array (DirectiveElaborator × Option String × Option SigDoc)) := do
```
david-christiansen, line 899: Shouldn't the function be renamed too?
```suggestion
private unsafe def directiveElabsForUnsafe' (x : Name) : DocElabM (Array (DirectiveElaborator × Option String × Option SigDoc)) := do
```

PR #699, opened 2026-01-13
test: LSP server verso tests

ejgallego:
We add LSP tests for Verso documents; we reuse the core LSP testing infrastructure from lean4 upstream. See `doc/dev/testing.md` there for more details.

This is a draft PR, some comments:

- maybe we should make the effort and use the lake test runner instead of the ad-hoc test runner here
- location of files / entry point to be debated
- we need to ship more tests

### DIFF
--- a/doc/UsersGuide/Releases/v4_28_0.lean
+++ b/doc/UsersGuide/Releases/v4_28_0.lean
@@ -18,4 +18,3 @@ file := "v4.28.0"
 %%%
 
 * Add Release Notes / Changelog to Verso Users guide (@ejgallego, #708)
-
--- a/src/tests/TestMain.lean
+++ b/src/tests/TestMain.lean
@@ -153,6 +153,16 @@ def testBlog (_ : Config) : IO Unit := do
   if fails > 0 then
     throw <| .userError s!"{fails} blog tests failed"
 
+-- Interactive tests via the LSP server
+def testInteractive (_ : Config) : IO Unit := do
+  IO.println "Running interactive (LSP) tests..."
+  IO.println s!"current dir: {(← IO.Process.getCurrentDir)}"
+  -- We use the lower-level Process.spawn, which causes the subprocess to inherit the stdio
+  let child ← IO.Process.spawn { cmd := "src/tests/interactive/run_interactive.sh" }
+  let exitCode ← child.wait
+  if exitCode != 0 then
+    throw <| IO.userError s!"Interactive LSP tests failed with exit code {exitCode}"
+
 open Verso.Integration in
 def tests := [
   testSerialization,
@@ -165,7 +175,8 @@ def tests := [
     (extraFiles := [("src/tests/integration/extra-files-doc/test-data/shared", "shared")])
     (extraFilesTeX := [("src/tests/integration/extra-files-doc/test-data/TeX-only", "TeX-only")]),
   testTexOutput "front-matter-doc" FrontMatter.doc,
-  testZip
+  testZip,
+  testInteractive
 ]
 
 def getConfig (config : Config) : List String → IO Config
--- a/src/tests/interactive/README.md
+++ b/src/tests/interactive/README.md
@@ -1,0 +1,38 @@
+# LSP Server Tests for Verso
+
+This directory contains the infrastructure and test cases for Verso
+LSP tests.
+
+See
+[Lean's LSP testing documentation](https://github.com/leanprover/lean4/blob/master/doc/dev/testing.md#test-suite-organization)
+for more information about how to test Lean's LSP server in Verso.
+
+## Adding test cases
+
+To add a test case, add two files to `test-cases` directory:
+
+- A `test.lean` file, with your test case according to the upstream
+  documentation.
+- A `test.lean.expected.out` that contains the expected test output.
+  This can be produced by first creating `test.lean`, running it, and
+  verifying that `test.lean.produced.out` has the expected content.
+  Then, `test.lean.produced.out` can be copied to
+  `test.lean.expected.out` and committed.
+
+## Running individual tests
+
+Do `./src/tests/interactive/run_single.sh $test_case` from Verso's
+root directory.
+
+## Runner architecture
+
+The runner architecture lies in a small custom script
+`./run_interactive.sh`. This script will call the upstream runner for
+each file in `test-cases`. The script is called from the main Verso
+test runner, in `src/tests/TestMain.lean`.
+
+Files from upstream:
+
+- `run.lean`: main runner
+- `test_single.sh`: test a single file
+- `common.sh`: diff and miscellaneous testing utilities
--- a/src/tests/interactive/test-cases/completion_inline.lean.expected.out
+++ b/src/tests/interactive/test-cases/completion_inline.lean.expected.out
@@ -1,0 +1,8 @@
+{"textDocument":
+ {"uri": "file:///src/tests/interactive/test-cases/completion_inline.lean"},
+ "position": {"line": 9, "character": 7}}
+{"items": [], "isIncomplete": true}
+{"textDocument":
+ {"uri": "file:///src/tests/interactive/test-cases/completion_inline.lean"},
+ "position": {"line": 13, "character": 18}}
+{"items": [], "isIncomplete": true}
--- a/.github/workflows/copyright-header.yml
+++ b/.github/workflows/copyright-header.yml
@@ -15,7 +15,7 @@ jobs:
 
             - name: Verify .lean files start with a copyright header.
               run: |
-                  FILES=$(find ./src ./test-projects -type f -name "*.lean" -exec perl -ne 'BEGIN { $/ = undef; } print "$ARGV\n" if !m{\A/-\nCopyright}; exit;' {} \;)
+                  FILES=$(find ./src ./test-projects -path ./src/tests/interactive -prune -type f -name "*.lean" -exec perl -ne 'BEGIN { $/ = undef; } print "$ARGV\n" if !m{\A/-\nCopyright}; exit;' {} \;)
                   if [ -n "$FILES" ]; then
                     echo "Found .lean files which do not have a copyright header:"
                     echo "$FILES"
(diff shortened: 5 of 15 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+unreleased
+----------
+
+(Changes that can break existing programs are marked with a "*")
```
david-christiansen: This might be confusing with Markdown already using `*` for bulleted lists and boldface. Maybe we should have two headers?

ejgallego:
Summary of PR review discussion with David:
- run test-runner via lake, maybe update test runner upstream to support `-d`
- make tests run from `lake test`
- note which files are copied from lean upstream
- test all Verso LSP extensions
- move changelog to manual, à la reference manual

CHANGELOG.md
ejgallego: We agree to follow what the ref manual does and just write "breaking change" in the changes entry.

david-christiansen:
> test all Verso LSP extensions

I think this one isn't really necessary for a merge here. It's important, but so is avoiding bitrot and getting value from the test that's already here. One or two more should suffice to validate the approach.

ejgallego:
>  update test runner upstream to support -d option

Upstream is happy with this so I'll proceed with that plan.

ejgallego:
> Summary of PR review discussion with David:
> 
> * run test-runner via lake, maybe update test runner upstream to support `-d`
> * make tests run from `lake test`
> * note which files are copied from lean upstream
> * test all Verso LSP extensions
> * move changelog to manual, à la reference manual

Points addressed, except for "test all Verso LSP extensions".

I added a custom pass on Verso test runner to run the tests, would be nice to have a fancier test runner at some point so we can run the tests in parallel for example.

doc/UsersGuide/Releases/v4_28_0.lean
```diff
@@ -18,4 +18,4 @@ file := "v4.28.0"
 %%%
 
 * Add Release Notes / Changelog to Verso Users guide (@ejgallego, #708)
-
+* Add LSP testing infrastructure for Verso documents (@ejgallego, #699)
```
david-christiansen: Are we using this for user-visible changes?

If we're going to include non-user-visible PRs, then perhaps we should have a process where we scrape GH like Lean uses rather than one where we manually do it? Or an action to ensure that the current PR is mentioned here as part of CI?

src/tests/interactive/test-cases/completion_inline.lean.expected.out
```diff
@@ -0,0 +1,8 @@
+{"textDocument":
+ {"uri": "file:///src/tests/interactive/test-cases/completion_inline.lean"},
+ "position": {"line": 9, "character": 7}}
+{"items": [], "isIncomplete": true}
+{"textDocument":
+ {"uri": "file:///src/tests/interactive/test-cases/completion_inline.lean"},
+ "position": {"line": 13, "character": 18}}
+{"items": [], "isIncomplete": true}
```
david-christiansen, line 8: These results seem incomplete. Is it just because you haven't merged #700 yet, and that will update the test?

src/tests/interactive/README.md
```diff
@@ -0,0 +1,34 @@
+## LSP Server Tests for Verso
```
david-christiansen:
```suggestion
# LSP Server Tests for Verso
```

src/tests/interactive/README.md
```diff
@@ -0,0 +1,34 @@
...
+[Lean's LSP testing documentation](https://github.com/leanprover/lean4/blob/master/doc/dev/testing.md#test-suite-organization)
+for more information about how to test Lean's LSP server in Verso.
+
+## Adding test cases
+
+To add a test case, add two files to `test-cases` directory:
+
+- A `test.lean` file, with your test case according to the upstream
+  documentation.
+- A `test.lean.expected.out` that contains the expected test output.
```
david-christiansen, line 16:
```suggestion
- A `test.lean.expected.out` that contains the expected test output. This can be produced by first creating `test.lean`, running it, and verifying that `test.lean.produced.out` has the expected content. Then, `test.lean.produced.out` can be copied to `test.lean.expected.out` and committed.
```

src/tests/TestMain.lean
```diff
@@ -153,6 +153,19 @@ def testBlog (_ : Config) : IO Unit := do
   if fails > 0 then
     throw <| .userError s!"{fails} blog tests failed"
 
+-- Interactive tests via the LSP server
+def testInteractive (_ : Config) : IO Unit := do
+  IO.println "Running interactive (LSP) tests..."
+  IO.println s!"current dir: {(← IO.Process.getCurrentDir)}"
+  -- We use the lower-level Process.spawn not to buffer the test output
+  -- Is this OK on windows ?
```
david-christiansen: Should we set up a Windows runner so you can be sure? No core Verso developer runs Windows, and all sorts of things are different over there; it'd be nice to know that we can build and test everything.

src/tests/TestMain.lean
```diff
@@ -153,6 +153,19 @@ def testBlog (_ : Config) : IO Unit := do
   if fails > 0 then
     throw <| .userError s!"{fails} blog tests failed"
 
+-- Interactive tests via the LSP server
+def testInteractive (_ : Config) : IO Unit := do
+  IO.println "Running interactive (LSP) tests..."
+  IO.println s!"current dir: {(← IO.Process.getCurrentDir)}"
+  -- We use the lower-level Process.spawn not to buffer the test output
```
david-christiansen:
```suggestion
  -- We use the lower-level Process.spawn, which causes the subprocess to inherit the stdio
  -- file descriptors from the parent, so its output is immediately visible during test runs.
```

src/tests/TestMain.lean
```diff
@@ -153,6 +153,19 @@ def testBlog (_ : Config) : IO Unit := do
...
+  -- Is this OK on windows ?
+  let child ← IO.Process.spawn { cmd := "src/tests/interactive/run_interactive.sh" }
+  let exitCode ← child.wait
+  if exitCode != 0 then
+    throw <| IO.userError s!"[!!!] interactive tests failed!"
```
david-christiansen:
```suggestion
    throw <| IO.userError s!"Interactive LSP tests failed with exit code {exitCode}"
```
I think the presence of an exception is exciting enough - we don't need the four exclamation marks.

src/tests/TestMain.lean
```diff
@@ -153,6 +153,19 @@ def testBlog (_ : Config) : IO Unit := do
...
+  -- Likely a more friendly rule on non-Unix
```
david-christiansen: Why not do this on Unix too?

I'd prefer we figure out the right thing to do here, and then do it, rather than including alternatives in comments. These have a tendency to stick around for a long time and impose a cognitive cost on readers who have to figure out the status of the comment's information each time they read the code.

If we consider an alternative and then reject it, documenting that reason in a comment is useful. Or, if one alternative is chosen arbitrarily, documenting that in a comment can also be useful in some cases. But I don't think this kind of comment is very useful for future readers.

doc/UsersGuide/Releases/v4_28_0.lean
ejgallego: Good point, I was indeed wondering whether we should include only user-visible changes in the log.

What do you think? I don't have a strong opinion myself.

src/tests/interactive/test-cases/completion_inline.lean.expected.out
ejgallego: Yes.

The idea here was to merge a test-case showcasing the bug.

I'm happy to push this test to the PR fixing the bug if we don't want to add this kind of tests.

src/tests/TestMain.lean
ejgallego: Happy to setup a windows runner, I think it is a key platform.

src/tests/TestMain.lean
ejgallego: IMHO the right fix here would be to use a Lean test runner instead of the `.sh` script.

I propose to keep the current call to the `.sh` script for now, instead of the test runner micro-managing shell execution. WDYT?

src/tests/TestMain.lean
ejgallego: See #724 for a (low-prio) attempt to add Windows CI; I propose to keep things as they are until the Windows env to run CI is settled down.

src/tests/interactive/test-cases/completion_inline.lean.expected.out
david-christiansen: This is fine, just wanted to confirm.

ejgallego:
Thanks for the reviews @david-christiansen , I've incorporated your feedback. Open points:
- add a changelog entry for this PR?
- keep tests that showcase a bug, or move them to the fix PR?
- are we happy with using `IO.Process.spawn { cmd := "src/tests/interactive/run_interactive.sh" }` for now?

doc/UsersGuide/Releases/v4_28_0.lean
david-christiansen: I think there's no great reason to duplicate information that's easy to find via `git log` or the GitHub web interface, and that this log makes sense as a way to communicate to users what we think is relevant to them. So I wouldn't include things like internal test improvements.

src/tests/TestMain.lean
david-christiansen: I think that's fine, I just don't think we should have comments that seem to imply we did something the wrong way. If we're in doubt, we should find out. Unless it doesn't matter, and then there's no need to find out or to have the comment.

doc/UsersGuide/Releases/v4_28_0.lean
ejgallego: Sounds good!

david-christiansen:
My opinions:

> * add a changelog entry for this PR?

No, we should save that for user-visible things.

> * keep tests that showcase a bug, or move them to the fix PR?

Keep 'em, and fix them in that PR :)

> * are we happy with using `IO.Process.spawn { cmd := "src/tests/interactive/run_interactive.sh" }` for now?

Yes

ejgallego:
Thanks, we should then be ready to go.

PR #700, opened 2026-01-13
fix: create well-spanned syntax on verso blocks

ejgallego:
We did push `genre` and `title` to the last block syntax, but this
broke the infotree span invariant, causing problems when retrieving
the goals.

Co-authored-by: David Thrane Christiansen <david@lean-fro.org>

PR #707, opened 2026-01-22
fix: Don't include syntax information in elaboration of math elements

ejgallego:
This avoids spurious infoview display of the internal elaborated
constant type.

This bug is only observable after applying the fix in #700 ; will wait
for #700 and #699 to merge this so we can properly provide a test.

Co-authored-by: David Thrane Christiansen <david@lean-fro.org>


ejgallego:
I think this is ready; note that I could only reproduce the bug for display math elements (first commit), but I guess fixing both situations is OK.

Using the LSP syntax `--^ rpcCall` worked for this case, but may not scale for more complex Verso tests.

PR #708, opened 2026-01-22
doc: Add changelog to users guide.

ejgallego:


doc/UsersGuide/Releases/v4_27_0.lean
```diff
@@ -0,0 +1,25 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Emilio J. Gallego Arias
+-/
+
+import VersoManual
+
+-- EJGA: Should we use the markdown code block from ref-manual?
```
david-christiansen: I think not. That exists as an artifact of the "export from GitHub" feature. We can just write them in Verso.

doc/UsersGuide/Releases.lean
```diff
@@ -0,0 +1,25 @@
...
+
+#doc (Manual) "Release Notes" =>
+%%%
+tag := "release-notes"
+file := "releases"
+number := false
+htmlSplit := .never
+%%%
+
+This section provides release notes about recent versions of Verso. When updating to a new version, please
```
david-christiansen, line 21: I think we should have a little note here that our version numbers are synced with the Lean version, so there's one Verso release per Lean release with no expectation of cross-Lean-version compatibility. In other words, our version number in this manual is not a semantic version.

ejgallego: Thanks for the comments. Should be addressed.

doc/UsersGuide/Releases/v4_27_0.lean
ejgallego: Thanks! Updated to:

```
* Add Release Notes / Changelog to Verso Users guide (@ejgallego, #708)
```

Would that work OK? I'd be happy to implement a PR role, so we could instead do `{pr 708}`.

doc/UsersGuide/Releases.lean
```diff
@@ -0,0 +1,27 @@
...
+read the corresponding release notes. They may contain advice that will help you understand
+the differences with the previous version and upgrade your projects.
+
+Note that Verso versioning follows Lean's, that is to say, we release a new version of version each time Lean releases, usually once per month. In particular, note that Verso doesn't follow the [semantic versioning model](https://semver.org/).
```
david-christiansen:
```suggestion
Verso versioning follows Lean's.
This means that we release a new version for each Lean release, usually once per month.
In particular, note that Verso doesn't follow the [semantic versioning model](https://semver.org/).
```

doc/UsersGuide/Releases/v4_27_0.lean
david-christiansen: I think `{pr}[708]` would be nice, but we don't need it. If you feel like it, go for it, but better to get the rest of your dependent PRs merged.

ejgallego: Makes sense, thanks. I'll do the `{pr ...}` role in a separate, low-prio PR.

PR #709, opened 2026-01-22
[PDF] increase feature coverage

jcreedcmu:


david-christiansen:
I think the diff on failed tests should be set up to show just a few lines of context, rather than the whole file. But this looks like it just needs an update-expected.

src/verso-manual/VersoManual.lean
```diff
@@ -91,9 +91,18 @@ inline_extension Inline.ref (canonicalName : String) (domain : Option Name) (rem
...
+    open Verso.Doc.TeX in
+    some <| fun go _ info content => do
+      match FromJson.fromJson? (α := RefInfo) info with
+      | .error e =>
+        TeX.logError e; content.mapM go
+      | .ok { canonicalName := name, domain, remote := none, resolvedDestination := none } =>
+        TeX.logError ("No destination found for tag '" ++ name ++ "' in " ++ toString domain); content.mapM go
+      | .ok { canonicalName := name, domain, remote := some remote, resolvedDestination := none } =>
+        TeX.logError ("No destination found for remote '" ++ remote ++ "' tag '" ++ name ++ "' in " ++ toString domain); content.mapM go
+      | .ok {resolvedDestination := some dest, ..} =>
```
david-christiansen: Do you think we can factor out the common logic here to a helper?

Something like:
```
where
  findDest logError go info := do
    match FromJson.fromJson? (...) info with
    | .error e => logError e; content.mapM go
    ...
```
I don't know if the monads can work out here, but it seems unfortunate to duplicate the JSON (de)serialization code - minor differences/bugs could easily sneak in, I think.

src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -134,6 +134,13 @@ private def andList (xs : Array Html) : Html :=
     open Html in
     (xs.extract 0 (xs.size - 1)).foldr (init := {{" and " {{xs.back!}} }}) (· ++ ", " ++ ·)
 
+private def andListTeX (xs : Array TeX) : TeX :=
+  if h : xs.size = 1 then xs[0]
+  else if h : xs.size = 2 then xs[0] ++ " and " ++ xs[1]
+  else
+    open TeX in
+    (xs.extract 0 (xs.size - 1)).foldr (init := \TeX{" and " \Lean{xs.back!} }) (· ++ ", " ++ ·)
```
david-christiansen: This function will fail if called with an empty array, but the error message will be a bit confusing (in terms of back! rather than andListTeX). What about adding `if h : xs.size = 0 then panic!"..."`, after which `back!` can probably become `back`?

david-christiansen: There should also be a docstring that describes when it will panic, I think.

The function could also take a proof that the array is nonempty.

src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -210,6 +266,43 @@ where
...
+    pure <| andListTeX (← ps.toArray.mapM (·.bibTeX go))
+where
+  authorTeX p := open TeX in do
+    if p.authors.size = 0 then
+      pure .empty
+    else if h : p.authors.size = 1 then
+      go <| Bibliography.lastName p.authors[0]
+    else if h : p.authors.size > 3 then
+      (· ++ \TeX{\em{"et al"} }) <$> go (Bibliography.lastName p.authors[0])
+    else andListTeX <$> p.authors.mapM (go ∘ Bibliography.lastName)
```
david-christiansen, line 312: Here you call it in a position where the proof should be automatic-ish, right?

src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -210,6 +266,43 @@ where
...
+    (fmt : Style) :
+    TeXT Manual (ReaderT ExtensionImpls IO) TeX := open TeX in do
+  match fmt with
+  | .textual =>
+    let out : Array TeX ← ps.toArray.mapM fun p => do
+      let m ← p.bibTeX go
+      pure <| \TeX{
+        \Lean{ ← authorTeX p} " "
+        \Lean{toString p.year}
+        \Lean{Marginalia.TeX m}
```
david-christiansen, line 290: Are marginal notes what we want for TeX output? I would think that we could either expand to a use of natbib-style citations here, or alternatively use TeX footnotes if we're slinging it by hand?

I guess the hybrid manual-work/bibtex approach seems to bring on the pain of both to me, without a clear benefit. What's the reasoning here?

src/verso-manual/VersoManual/License.lean
```diff
@@ -75,12 +75,42 @@ where
           .empty
       {{<section>{{hdrHtml}}{{paragraphedHtml txt}}</section>}}
 
+open Verso.Output.TeX in
+def LicenseInfo.toTeX [Monad m] [Doc.TeX.GenreTeX g m] (license : LicenseInfo) (headerLevel : Nat) :
```
david-christiansen: Do we need this?

The reason for LicenseInfo is that the rendered HTML contains code copyright/licensed by others (e.g. the hover box JavaScript), and various licenses require acknowledgment. But I don't think the TeX output should be including a license notice for e.g. KaTeX when it's not using it in any way, and TeX dependencies aren't bundled into the document in the same way that JS code is into the HTML.

What if the TeX version of this just emitted nothing, and the docs were updated to state that the feature is for the HTML backend?

src/verso-manual/VersoManual/License.lean
```diff
@@ -75,12 +75,42 @@ where
...
 block_extension Block.licenseInfo where
   traverse _ _ _ := do
     pure none
-  toTeX := some <| fun _ _ _ _ _ => pure .empty
+  toTeX := open Verso.Output.TeX in
+    some <| fun _ _ _ _ _ => do
+      let ⟨_, ctx, state, _⟩ := (← read)
+      let headerLevel := ctx.headers.size + 1
+      let allLicenses := state.licenseInfo.toArray
+      let allLicenses := allLicenses.qsort (·.dependency.trimAscii.copy.toLower < ·.dependency.trimAscii.copy.toLower)
```
david-christiansen: This is probably fine, but it does seem to profligately allocate strings :)

What about
```
let allLicenses := allLicenses.map (fun x => (x, x.trimAscii.copy.toLower)) |>.qsort (·.2 < ·.2) |>.map (·.1)
```
?

src/verso/Verso/Doc/TeX.lean
```diff
@@ -47,6 +47,14 @@ public def texContext [Monad m] : TeXT g m TeXContext := do
   let ⟨_, _, _, ctx⟩ ← read
   pure ctx
 
+-- FIXME: would like `header` to call this, but how to best
+-- streamline error handling to avoid making two redundant checks?
```
david-christiansen: What about making it a private function that's only called after the error is handled? Maybe in a `where` block?

david-christiansen: Also it looks like you're not calling this...

src/verso-manual/VersoManual/Bibliography.lean
david-christiansen: Decision from call: we'll drop the BibTeX, render cites as footnotes for now. We can revisit if it matters in the future.

src/verso-manual/VersoManual/Bibliography.lean
jcreedcmu: Ah, I agree having the panic in an explicit if-empty branch is better, adding that to both HTML and TeX.

src/verso-manual/VersoManual/Bibliography.lean
jcreedcmu: In that position, yes, but there are a handful of other callsites where the data comes from higher up and isn't obviously validated to be nonempty. Could try to bake that constraint into types, but leaving that outside the scope of this PR.

src/verso-manual/VersoManual/License.lean
jcreedcmu: Agree after discussion that emitting nothing is ok

src/verso-manual/VersoManual/License.lean
jcreedcmu: moot because of decsiion to omit license information.

src/verso/Verso/Doc/TeX.lean
jcreedcmu: It is in fact used in `/src/verso-manual/VersoManual/Literate.lean` but I'm not entirely sure why `find-references` *doesn't* list this use for me. I only find out once I try to compile and fail.

src/verso-manual/VersoManual.lean
jcreedcmu: Agree this is a fine idea across the board, but leaving it out of this PR.

src/verso/Verso/Doc/TeX.lean
jcreedcmu: A moment's thought led to a way of splitting out the actual error message, but I was slightly surprised not to find something that might have been reasonably named
`Fin.mk? : (a n : ℕ) : Option (Fin n)`
with the property that if `a < n`, then `Fin.mk a n = some (Fin.mk ⟨a, h⟩)` for some
`h`, and `none` otherwise. Such a thing being in the prelude or whatever would have been nice to golf what I wrote a tiny bit.

PR #727, opened 2026-02-03
doc: add link to rendered manual

david-christiansen:


david-christiansen:
@ejgallego Do you think this is the right place for the link?

ejgallego:
> @ejgallego Do you think this is the right place for the link?

It is where I would expect it, yes.

I guess we could specify that the currently deployed version belongs to the tip of the `main` branch? But otherwise I think this is fine.

ejgallego:
By the way, should #725 get a changelog entry?

david-christiansen:
I wouldn't think so. But I could be convinced! I see it as a dev change rather than a user one, because it's all things based on going to the source repository.

PR #734, opened 2026-02-05
chore: enable Dependabot for GitHub Actions

david-christiansen:
Let's see if this reduces the work to tread water here.


david-christiansen:
!prettier

PR #763, opened 2026-02-23
feat: improve role resolution error diagnostics

ejgallego:
This PR improves diagnostics for Verso document extension resolution, starting from inline role resolution and now sharing the same diagnostic machinery with code blocks, directives, and block commands where appropriate.

It reports explicit extension-resolution failures for:

- a declaration that can be used as an expander but is not registered for the requested extension syntax
- a declaration that exists but is not registered for the requested extension syntax
- an unknown extension name, with edit-distance-ranked clickable hints when there are close matches and no broad fallback list when the name is not close enough

The PR also:

- preserves existing block-command fallback to ordinary declarations when the name resolves but is not registered as a block command
- exposes registered extension names for diagnostics across roles, code blocks, directives, and block commands
- uses shadowing-aware display names for suggestions
- factors shared extension expander execution in `Verso.Doc.Elab`
- keeps the new `VersoManual.Index` role imports scoped to the public meta inline elaboration dependency
- registers legacy inline APIs as explicit roles for compatibility (`today`, `date`, `sectionRef`, `index`, `see`, `seeAlso`)
- adds `Tests.ExtensionResolution`, including explicit distance-matching and no-match cutoff coverage
- updates the current release notes

Follow-up work remains appropriate for role/directive/command completion.

This PR was prepared with help from Codex (model: GPT 5.5)

### DIFF
--- a/src/tests/Tests/ExtensionResolution.lean
+++ b/src/tests/Tests/ExtensionResolution.lean
@@ -110,10 +110,39 @@ end RoleCases
 
 namespace DistanceCases
 
+/-
+The suggestion cutoff table is:
+* input length less than 3: Levenshtein distance at most 1
+* input length less than 6: Levenshtein distance at most 2
+* otherwise: Levenshtein distance at most 3
+
+The length-1 no-match case appears before this namespace declares a local single-character role,
+because any single-character role is within distance 1 of any single-character typo.
+-/
+
+/--
+error: No registered role `q`.
+-/
+#guard_msgs in
+#docs (.none) oneCharDistanceNoMatch "One-character distance no match" :=
+:::::::
+{q}[]
+:::::::
+
+@[role]
+def r : RoleExpanderOf Unit
+  | (), _ => do
+    `(Verso.Doc.Inline.text "one-character-cutoff-role")
+
 @[role]
 def ab : RoleExpanderOf Unit
   | (), _ => do
     `(Verso.Doc.Inline.text "short-distance-role")
+
+@[role]
+def vvv : RoleExpanderOf Unit
+  | (), _ => do
+    `(Verso.Doc.Inline.text "three-character-cutoff-role")
 
 @[role]
 def distanceRegistered : RoleExpanderOf Unit
@@ -141,6 +170,18 @@ def multiAlphi : RoleExpanderOf Unit
     `(Verso.Doc.Inline.text "multi-alphi-role")
 
 /--
+error: No registered role `q`.
+
+Hint: Did you mean role `r`?
+  q̵r̲
+-/
+#guard_msgs in
+#docs (.none) oneCharDistanceMatch "One-character distance match" :=
+:::::::
+{q}[]
+:::::::
+
+/--
 error: No registered role `ac`.
 
 Hint: Did you mean role `ab`?
@@ -161,12 +202,26 @@ error: No registered role `zz`.
 {zz}[]
 :::::::
 
-/-
-The suggestion cutoff table is:
-* input length less than 3: Levenshtein distance at most 1
-* input length less than 6: Levenshtein distance at most 2
-* otherwise: Levenshtein distance at most 3
+/--
+error: No registered role `vww`.
+
+Hint: Did you mean role `vvv`?
+  vw̵w̵v̲v̲
 -/
+#guard_msgs in
+#docs (.none) threeCharDistanceMatch "Three-character distance match" :=
+:::::::
+{vww}[]
+:::::::
+
+/--
+error: No registered role `www`.
+-/
+#guard_msgs in
+#docs (.none) threeCharDistanceNoMatch "Three-character distance no match" :=
+:::::::
+{www}[]
+:::::::
 
 /--
 error: No registered role `yyyxx`.
--- a/src/tests/Tests/ExtensionResolution.lean
+++ b/src/tests/Tests/ExtensionResolution.lean
@@ -130,6 +130,16 @@ def zzzzzz : RoleExpanderOf Unit
   | (), _ => do
     `(Verso.Doc.Inline.text "long-cutoff-role")
 
+@[role]
+def multiAlpha : RoleExpanderOf Unit
+  | (), _ => do
+    `(Verso.Doc.Inline.text "multi-alpha-role")
+
+@[role]
+def multiAlphi : RoleExpanderOf Unit
+  | (), _ => do
+    `(Verso.Doc.Inline.text "multi-alphi-role")
+
 /--
 error: No registered role `ac`.
 
@@ -198,6 +208,19 @@ error: No registered role `zzaaaa`.
 #docs (.none) longBoundaryDistanceNoMatch "Long boundary distance no match" :=
 :::::::
 {zzaaaa}[]
+:::::::
+
+/--
+error: No registered role `multiAlphx`.
+
+Hint: Did you mean role `multiAlpha`?
+  • multiAlphx̵a̲
+  • multiAlphx̵i̲
+-/
+#guard_msgs in
+#docs (.none) multiDistanceSuggestions "Multiple distance suggestions" :=
+:::::::
+{multiAlphx}[]
 :::::::
 
 /--
--- a/src/tests/Tests/ExtensionResolution.lean
+++ b/src/tests/Tests/ExtensionResolution.lean
@@ -108,6 +108,62 @@ error: No registered role `nothereatallzzzz`.
 
 end RoleCases
 
+namespace DistanceCases
+
+@[role]
+def ab : RoleExpanderOf Unit
+  | (), _ => do
+    `(Verso.Doc.Inline.text "short-distance-role")
+
+@[role]
+def distanceRegistered : RoleExpanderOf Unit
+  | (), _ => do
+    `(Verso.Doc.Inline.text "distance-role")
+
+/--
+error: No registered role `ac`.
+
+Hint: Did you mean role `ab`?
+  ac̵b̲
+-/
+#guard_msgs in
+#docs (.none) shortDistanceMatch "Short distance match" :=
+:::::::
+{ac}[]
+:::::::
+
+/--
+error: No registered role `zz`.
+-/
+#guard_msgs in
+#docs (.none) shortDistanceNoMatch "Short distance no match" :=
+:::::::
+{zz}[]
+:::::::
+
+/--
+error: No registered role `distanceRegistred`.
+
+Hint: Did you mean role `distanceRegistered`?
+  distanceRegiste̲red
+-/
+#guard_msgs in
+#docs (.none) longDistanceMatch "Long distance match" :=
+:::::::
+{distanceRegistred}[]
+:::::::
+
+/--
+error: No registered role `distanceNoMatchzzzz`.
+-/
+#guard_msgs in
+#docs (.none) longDistanceNoMatch "Long distance no match" :=
+:::::::
+{distanceNoMatchzzzz}[]
+:::::::
+
+end DistanceCases
+
 namespace ShadowSource
 
 @[role]
--- a/src/tests/Tests/ExtensionResolution.lean
+++ b/src/tests/Tests/ExtensionResolution.lean
@@ -120,6 +120,16 @@ def distanceRegistered : RoleExpanderOf Unit
   | (), _ => do
     `(Verso.Doc.Inline.text "distance-role")
 
+@[role]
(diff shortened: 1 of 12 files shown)


src/tests/Tests/RoleResolution.lean
```diff
@@ -0,0 +1,171 @@
...
+error: No registered role `registred`. Did you mean role `registered`?
+-/
+#guard_msgs in
+#docs (.none) roleCase4 "Case 4" :=
+:::::::
+{registred}[]
+:::::::
+
+/--
+error: No registered role `nothereatallzzzz`. Available roles (showing 20 of 37): anchor, anchorError, anchorInfo, anchorName, anchorTerm, anchorWarning, blob, citehere, citep, citet, conv, deftech, draft, htmlSpan, inst, label, lean, leanCommand, leanInline, leanKw
```
david-christiansen: This would be nicer if it sorted the suggestions by Levenshtein distance from what the user wrote, with a cutoff depending on the length of the string (maybe threshold of 1 if length < 3, 2 if length < 7, 3 otherwise?).

And it should use the hint mechanism to make it clickable.

ejgallego: Implemented as requested:
- Suggestions are ranked by Levenshtein distance with cutoff based on input length (`<3 => 1`, `<7 => 2`, else `3`).
- Suggestions are emitted via the hint mechanism so they are clickable.

Follow-up refinement: for the no-close-match fallback, hints now show only the 5 closest roles to avoid overwhelming output.

ejgallego:
@david-christiansen ready for review IMO.

src/tests/Tests/RoleResolution.lean
```diff
@@ -0,0 +1,201 @@
...
+/-- info: #[Verso.Doc.Block.para #[Verso.Doc.Inline.text "registered-role"]] -/
+#guard_msgs in
+#eval roleCase1.toPart.content
+
+def unregistered : Verso.Doc.Elab.RoleExpander
+  | _, _ => do
+    pure #[← `(Verso.Doc.Inline.text "unregistered-role")]
+
+/--
+error: Role function `unregistered` was found but not registered as a role. Register it with `@[role]` or `@[role_expander ...]`.
```
david-christiansen:
```suggestion
error: Role function `unregistered` was found but not registered as a role. Register it with `@[role]`.
```

I think the `role_expander` attribute should be considered legacy at this point.

src/tests/Tests/RoleResolution.lean
```diff
@@ -0,0 +1,201 @@
...
+#guard_msgs in
+#docs (.none) roleCase2 "Case 2" :=
+:::::::
+{unregistered}[]
+:::::::
+
+def wrongType : Nat := 7
+
+/--
+error: Function `wrongType` was found but likely not a role.
```
david-christiansen: As a nitpick, these are not necessarily functions :-)  A role can be attached to any Lean constant; by convention, we use the expander itself when nothing else is needed, but there are cases where we need to attach it to something else.

src/tests/Tests/ExtensionResolution.lean
```diff
@@ -0,0 +1,201 @@
...
+#docs (.none) roleCase3 "Case 3" :=
+:::::::
+{wrongType}[]
+:::::::
+
+/--
+error: No registered role `registred`.
+
+Hint: Did you mean role `registered`?
+  registe̲red
```
david-christiansen, line 80: This is great!

src/tests/Tests/RoleResolution.lean
```diff
@@ -0,0 +1,201 @@
...
+
+/--
+error: No registered role `nothereatallzzzz`.
+
+Hint: Closest registered roles:
+  • n̵o̵t̵h̵e̵r̵e̵a̵t̵a̵l̵l̵z̵z̵z̵z̵c̲i̲t̲e̲h̲e̲r̲e̲
+  • n̵o̵t̵h̵e̵r̵e̵a̵t̵a̵l̵l̵z̵z̵z̵z̵a̲n̲c̲h̲o̲r̲N̲a̲m̲e̲
+  • n̵o̵t̵h̵e̵r̵e̵a̵t̵a̵l̵l̵z̵z̵z̵z̵a̲n̲c̲h̲o̲r̲W̲a̲r̲n̲i̲n̲g̲
+  • n̵o̵t̵h̵e̵r̵e̵a̵t̵a̵l̵l̵z̵z̵z̵z̵c̲i̲t̲e̲t̲
+  • n̵o̵t̵h̵e̵r̵e̵a̵t̵a̵l̵l̵z̵z̵z̵z̵h̲t̲m̲l̲S̲p̲a̲n̲
```
david-christiansen: This message strikes me as confusing. Perhaps we need a cutoff on the Levenshtein distance? Like (max 2 (roleName.toString.length / 6) ? Or perhaps even just a table where length < 3 gives threshold 1, length < 6 gives 2, and otherwise 3 ?

david-christiansen: This test will also be very unstable as Verso evolves

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -627,6 +627,19 @@ private unsafe def roleExpandersForUnsafe (x : Name) : DocElabM (Array (RoleExpa
 @[implemented_by roleExpandersForUnsafe]
 public opaque roleExpandersFor (x : Name) : DocElabM (Array (RoleExpander × Option String × Option SigDoc))
 
+private unsafe def registeredRoleNamesUnsafe : DocElabM (Array Name) := do
```
david-christiansen: Why does this need to be unsafe? I don't see any eval going on here.

src/verso/Verso/Doc/Elab.lean
```diff
@@ -118,20 +119,108 @@ private meta def expanderDocHover (stx : Syntax) (what : String) (name : Name) (
     out := out ++ "\n\n" ++ d
   Hover.addCustomHover stx out
 
+private meta def roleSuggestionThreshold (input _candidate : String) : Nat :=
+  if input.length < 3 then 1 else if input.length < 7 then 2 else 3
```
david-christiansen: Ah, this is here. Odd that it didn't work in the test suite...

src/verso/Verso/Doc/Elab.lean
```diff
@@ -118,20 +119,108 @@ private meta def expanderDocHover (stx : Syntax) (what : String) (name : Name) (
...
+      hintSuggestions
+      (ref? := some name) (forceList := suggestions.size > 1)
+    throwErrorAt name m!"No registered role `{name.getId}`.{hint}"
+  | [] =>
+    if available.isEmpty then
+      throwErrorAt name m!"No registered role `{name.getId}`. No roles are currently registered."
+    else
+      let shown := closestRoleNames available requested
+      let hintSuggestions := shown.map fun (_, roleName) =>
+        ({suggestion := .string roleName} : Lean.Meta.Hint.Suggestion)
```
david-christiansen: This is not correct in the presence of shadowing. Instead, this should use one of the `unresolveName` functions, and it'd be good to test that it works as well by shadowing a role and looking at the suggestion.

src/verso/Verso/Doc/Elab.lean
```diff
@@ -118,20 +119,108 @@ private meta def expanderDocHover (stx : Syntax) (what : String) (name : Name) (
...
+
+private meta def throwUnknownRoleError (name : Ident) : DocElabM α := do
+  let requested := name.getId.toString
+  let available ← availableRoleDisplayNames
+  let suggestions := roleSuggestions available requested
+  match suggestions.toList with
+  | _ :: _ =>
+    let best := suggestions[0]!.2
+    let hintSuggestions := suggestions.map fun (_, roleName) =>
+      ({suggestion := .string roleName} : Lean.Meta.Hint.Suggestion)
```
david-christiansen: See below comment about shadowing

src/verso/Verso/Doc/Elab.lean
```diff
@@ -118,20 +119,108 @@ private meta def expanderDocHover (stx : Syntax) (what : String) (name : Name) (
...
+      let c ← mkConstWithLevelParams declName
+      discard <| Meta.mkAppM ``toRole #[c]
+      return true
+    catch
+      | _ => return false
+
+private meta def throwRoleNotRegisteredError (name : Ident) (resolvedName : Name) : DocElabM α := do
+  let shownName := shortRoleName resolvedName
+  if ← isRoleFunctionType resolvedName then
+    throwErrorAt name m!"Role function `{shownName}` was found but not registered as a role. Register it with `@[role]` or `@[role_expander ...]`."
```
david-christiansen:
```suggestion
    throwErrorAt name m!"Role function `{shownName}` was found but not registered as a role. Register it with `@[role]`."
```

src/tests/Tests/RoleResolution.lean
```diff
@@ -0,0 +1,201 @@
...
+-/
+#guard_msgs in
+#docs (Blog.Post) roleCase4 "Blog Case 4" :=
+:::::::
+{blogNoCloseMatchZzzzz}[]
+:::::::
+
+end BlogCases
+
+end Verso.RoleResolutionTest
```
david-christiansen:
```suggestion
```

david-christiansen:
Do you plan on implementing the other extension types in this PR? I think that makes the most sense from the perspective of keeping everything consistent, and it should be pretty easy once the role version is nailed down.

src/tests/Tests/RoleResolution.lean
ejgallego: Fixed thanks.

src/tests/Tests/RoleResolution.lean
ejgallego: Fixed thanks.

src/verso/Verso/Doc/Elab/Monad.lean
ejgallego: Good catch, fixed thanks.

src/verso/Verso/Doc/Elab.lean
ejgallego: Improved, test added.

ejgallego:
Thanks for the careful review David, I've done a non-trivial pass over the PR: Added coverage for other Verso constructions, improved tests, and refactored common Elab code.

src/tests/Tests/RoleResolution.lean
ejgallego: Improved tests, hopefully the new setup will be stable.

src/verso/Verso/Doc/Elab/ExtensionResolution.lean
```diff
@@ -0,0 +1,143 @@
...
+    try
+      return ← run expander
+    catch
+      | ex@(.internal id) =>
+        if id == unsupportedSyntaxExceptionId then pure ()
+        else throw ex
+      | ex => throw ex
+  throwUnsupportedSyntax
+
+end Verso.Doc.Elab
```
david-christiansen:
```suggestion
```

src/verso/Verso/Doc/Elab.lean
```diff
@@ -118,33 +119,37 @@ private meta def expanderDocHover (stx : Syntax) (what : String) (name : Name) (
...
+  tryExtensionExpanders expanders fun (e, doc?, sig?) => do
+    let termStxs ← withFreshMacroScope <| run e
+    expanderDocHover nameStx what resolvedName doc? sig?
+    match shape with
+    | .inline =>
+      let termStxs ← termStxs.mapM fun t => (``(($t : Inline $(⟨genre⟩))))
+      if h : termStxs.size = 1 then return termStxs[0]
+      else return (← ``(Inline.concat (genre := $(⟨genre⟩)) #[$[$termStxs],*]))
+    | .block =>
+      return (← ``(Block.concat (genre := $(⟨genre⟩)) #[$[$termStxs],*]))
```
david-christiansen, line 142: This refactoring is very nice, thanks!

src/tests/Tests/ExtensionResolution.lean
```diff
@@ -0,0 +1,508 @@
...
+-/
+#guard_msgs in
+#docs (.none) blockCommandTypo "Block command typo" :=
+:::::::
+{registeredComand}
+:::::::
+
+end BlockCommandCases
+
+end Verso.ExtensionResolutionTest
```
david-christiansen:
```suggestion
```

david-christiansen (approved):

PR #765, opened 2026-02-24
chore: expose PR preview as GitHub deployment

ejgallego:
This PR adds GitHub Deployment reporting to the pull-request manual preview workflow.

After PR CI builds and uploads the manual artifact, the follow-up deploy workflow publishes that artifact to Netlify and records the resulting preview URL as a transient GitHub Deployment for the PR preview environment. The deployment is attached to the PR workflow target SHA, with the PR head SHA as a fallback.

During rollout, the workflow also posts the familiar PR preview link as a new PR conversation comment. This keeps the existing reviewer experience in place while we verify that GitHub surfaces the deployment link reliably on PRs. Once that is confirmed, the comment step and its `issues: write` permission can be removed.

The workflow records the deployment explicitly because it runs under `workflow_run`. In that context, `nwtgck/actions-netlify` sees the default-branch workflow SHA as `context.sha` and does not expose a deployment ref/SHA override. The Netlify action is still used for the Netlify upload itself, but GitHub deployment metadata is recorded by a separate first-party `actions/github-script` step.

Supply-chain changes in the secret-bearing deploy workflow:

- PR metadata lookup no longer uses `potiuk/get-workflow-origin`; it now uses `actions/github-script@v9`.
- Artifact download no longer uses `dawidd6/action-download-artifact`; it now uses `actions/download-artifact@v7`.
- PR commenting no longer uses `marocchino/sticky-pull-request-comment`; it now uses `actions/github-script@v9`.
- The Netlify upload still uses `nwtgck/actions-netlify`, pinned to a full commit SHA, because that action receives `NETLIFY_AUTH_TOKEN`.
- The workflow no longer passes `GITHUB_TOKEN` to the Netlify action.

The deploy job runs after PR CI succeeds and deploys the uploaded artifact without checking out or executing PR code. A malicious PR can control the preview HTML content, as expected for a preview, but this workflow does not expose deploy secrets to the PR CI job.

Prepared with Codex (GPT-5.5)

### DIFF
--- a/.github/workflows/pr-deploy.yml
+++ b/.github/workflows/pr-deploy.yml
@@ -6,6 +6,13 @@ on:
         types:
             - completed
 
+permissions:
+    actions: read
+    contents: read
+    deployments: write
+    issues: write
+    pull-requests: read
+
 jobs:
     deploy:
         name: Deploy PR preview to Netlify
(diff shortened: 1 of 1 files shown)


david-christiansen:
This would have made it more difficult to review the dark mode PR (#641), for example.

What's the major benefit here? How do the comments hurt anything?

ejgallego:
> This would have made it more difficult to review the dark mode PR (#641), for example.

I agree.

> What's the major benefit here? How do the comments hurt anything?

The main source of noise is that I now get a comment notification for all PRs in the GitHub web interface due to the bot.

Unfortunately there seems no way to filter those.

david-christiansen:
What if we rejigger the CI so that it creates a deployment instead of posting a comment? If we can do that, then it may solve the issue.

ejgallego:
> What if we rejigger the CI so that it creates a deployment instead of posting a comment? If we can do that, then it may solve the issue.

I didn't know about GH deployment, that seems a great idea, thanks!

ejgallego:
I updated the branch to use a Github deployment. I don't have experience with those, it seems that testing them requires first merging this PR.

In theory, after we merge this PR, next PR preview CI action should create a deployment.

david-christiansen:
This looks plausible to me but I don't really have the background to review and understand it.

Perhaps we should set it up to do both the comment and the deployment, merge it, and then if it seems to be good for a few days, delete the comment? That we we don't have a regression while working.

ejgallego:
Hi @david-christiansen , PR updated in the lines we discussed:
- we keep the comment for now, but we also trigger a deployment
- we have to create the deployment manually, because the deployment script runs in the workflow_run context. netlify action has support to use github deployments, but it is not safe for pull requests due to secrets.
- I've reduced our supply-chain exposure to 3rd party github actions. In some cases, they were not needed, and in others, they could be replaced by simple scripts.

I'll go ahead and merge soon as it is the only way to test.

PR #766, opened 2026-02-24
fix: align inline Lean role names with Manual

### LINKED ISSUE #762: Inline {lean}`...` does not work
I'm writing a blog post on `v4.28.0` and the inline lean code does not work. Writing ``{lean}`Nat` `` produces:
```
Application type mismatch: The argument
  #[Doc.Inline.code "Nat"]
has type
  Array (Doc.Inline ?m.15)
but is expected to have type
  LeanBlockConfig
in the application
  lean #[Doc.Inline.code "Nat"]
```

Also, there used to be `` {name}`Nat` ``, is that gone? (I used Verso more than a year ago, so I'm not up to speed with recent changes)

ejgallego:
This PR harmonizes the naming of the inline Lean role between the blog and manual genres, so that users don't need to remember the difference.

Fixes #762

### DIFF
--- a/src/verso-blog/VersoBlog.lean
+++ b/src/verso-blog/VersoBlog.lean
@@ -599,8 +599,7 @@ where
     modifyInfoState fun s => { s with trees := f s.trees }
 
 open SubVerso.Highlighting Highlighted in
-@[role]
-def leanInline : RoleExpanderOf LeanInlineConfig
+private def leanInlineImpl : RoleExpanderOf LeanInlineConfig
   | config, elts => withTraceNode `Elab.Verso.block.lean (fun _ => pure m!"lean block") <| do
     let #[code] := elts
       | throwError "Expected precisely one code element"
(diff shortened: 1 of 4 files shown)


david-christiansen:
The PR title doesn't really align with the Lean conventions, and I think this new format is perhaps less useful in the Git log.

What about something like:

"This PR harmonizes the naming of the inline Lean role between the blog and manual genres, so that users don't need to remember the difference"?

at the start? The bullet points all seem to follow from here. And I don't think the specific list of test commands is so useful here, because our CI should be doing all those things, right? And some of them imply the others?

ejgallego:
> The PR title doesn't really align with the Lean conventions, and I think this new format is perhaps less useful in the Git log.
> 
> What about something like:
> 
> "This PR harmonizes the naming of the inline Lean role between the blog and manual genres, so that users don't need to remember the difference"?
> 
> at the start? The bullet points all seem to follow from here. And I don't think the specific list of test commands is so useful here, because our CI should be doing all those things, right? And some of them imply the others?

Yes, will fix. Sorry for the noise.

I told Codex to submit the PR for me and it added the validation comments and refined the PR title from my prompt.

I'm fixing my setup so that doesn't happen again.

ejgallego:
Ok, PR should be good for review now, sorry for the noise.

src/verso-blog/VersoBlog.lean
```diff
@@ -675,6 +674,19 @@ def leanInline : RoleExpanderOf LeanInlineConfig
...
 
+@[role lean]
+def leanCanonical : RoleExpanderOf LeanInlineConfig :=
+  leanInlineImpl
+
+@[role]
+def leanInline : RoleExpanderOf LeanInlineConfig
+  | config, elts => do
+    if h : 0 < elts.size then
+      logWarningAt elts[0] "`{leanInline}` is deprecated; use `{lean}` instead."
```
david-christiansen, line 685: Nice fix :-)

david-christiansen:
> Ok, PR should be good for review now, sorry for the noise.

No worries :-) I set up some automation to catch some of these things, so we don't have to worry about it.

src/verso-blog/VersoBlog.lean
ejgallego: Codex suggested that :-) :robot: :tophat: :clap: :clap:

PR #768, opened 2026-02-26
fix: Verso folding ranges / TOC for Lean.Doc syntax and #doc

ejgallego:
We fix table of contents (and by extension folding) for `#doc` syntax and users of `Lean.Doc` syntax (as opposed to `Verso.Doc`)

Problem 1 is that we forgot to push the TOC custom info at the end of `#doc` elaboration.

Problem 2 was due to the match in the LSP chain not being aware of the upstreaming of Verso parser. Maybe we should remove the aliases? Breakage can be fixed automatically these days.

This patch makes folding and TOC work as expected. Test included.

I am grouping the two fixes due to the test including both.

There is a little artifact related to #doc in the TOC (added twice) that I will fix in a follow up commit.

### DIFF
--- a/doc/UsersGuide/Releases.lean
+++ b/doc/UsersGuide/Releases.lean
@@ -6,6 +6,7 @@ Author: Emilio J. Gallego Arias
 
 import VersoManual
 
+import UsersGuide.Releases.«v4_29_0_rc2»
 import UsersGuide.Releases.«v4_28_0»
 
 open Verso Genre Manual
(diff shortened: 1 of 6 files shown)


david-christiansen:
Looks great - thanks!

ejgallego:
Thanks David!

Should should we remove the Verso.Doc aliases?

ejgallego:
Folding ranges and outline before the patch:

Before

Folding ranges and outline after the patch:

After

ejgallego:
> Folding ranges and outline before the patch:
> 
> Before
> 
> Folding ranges and outline after the patch:
> 
> After

This is an attempt to have Codex automate taking pictures of before / after changes for both Verso output and VSCode windows. After this little test, I'm positive we can make it work.

IMHO worth doing for PR review.

PR #769, opened 2026-02-26
chore: Bump toolchain to v4.29.0-rc2

ejgallego:
chore: Bump toolchain to v4.29.0-rc2


ejgallego:
@david-christiansen the failure on the About.lean is real. Investigating.

david-christiansen:
Me too! It looks like some kind of change in macro hygiene, but it really surprises me that this would change between RCs

david-christiansen:
I think that #774 superseded this one.  Thanks for investigating!

PR #770, opened 2026-02-26
[don't merge] demonstrate baffling grind failure

robsimmons:


david-christiansen:
It sounds like this issue got resolved - can we close this?

PR #771, opened 2026-02-26
fix: preserve Verso document ranges for LSP

ejgallego:
This PR fixes Verso's LSP document-symbol and folding ranges for document command syntax and term-position Verso documents.

Previously, root document parts did not preserve distinct syntax anchors for the full document range and the visible title selection. This could make a top-level document symbol start at the wrong position, merge unrelated Lean declarations under the document symbol, or produce a `selectionRange` outside its enclosing `range`. LSP clients such as VS Code reject the whole document-symbol response when that containment invariant is violated, which makes VSCode not to display the outline.

The fix stores separate syntax anchors for each part: `rangeSyntax` for the full document-symbol and folding range, and `selectionSyntax` for the symbol selection range. Root document elaboration captures the current elaborator reference before title elaboration as the range anchor while keeping the title literal as the selection anchor.

TOC construction now validates the syntax anchors before they reach the LSP handlers. This validation only checks the metadata that Verso is about to store: both anchors must have ranges recoverable by Lean's syntax APIs, and the selection anchor must be contained in the full part span from the range anchor start to the part end. The LSP layer then calls Lean's standard `FileMap.lspRangeOfStx?` helper directly, rather than carrying a Verso-specific range fallback or normalization path; TOC LSP conversion now panics if that previously-checked invariant is violated.

Explicit `PointOfInterest` selection syntax is also validated when the metadata is saved, so document-symbol construction consumes already-checked selection ranges instead of re-checking containment in the LSP handler.

For direct syntax, the important case is header synthetic nodes. Header wrapper syntax can have synthetic or missing source information even when its children retain recoverable positions; if Verso stores such a wrapper as the TOC range anchor without preserving a recoverable full span, document-symbol and folding extraction can later produce invalid or missing ranges. This PR keeps that provenance check at TOC creation, and `#docs` also uses the final token's tail position for document ends.

Finally, the term-position `#doc` elaborator now matches `elabDoc`'s `VersoDoc` result instead of expecting `Part`, and a small `Tests.DocTerm` smoke test covers direct term-`#doc` elaboration outside the LSP harness.

Prepared with Codex (GPT-5.5)


robsimmons (approved):
Heck yes. I wonder if there are reasons where/why it would be helpful to track the location of the `#doc` tag itself?

ejgallego:
> Heck yes. I wonder if there are reasons where/why it would be helpful to track the location of the `#doc` tag itself?

Indeed I wonder what'd be best here. I just had a look at the test output and indeed it still looks a bit suspicious. Would love to pair and brainstorm any time on this!

src/tests/interactive/test-cases/symbols_verso.lean.expected.out
```diff
@@ -0,0 +1,28 @@
...
+ {"uri": "file:///src/tests/interactive/test-cases/symbols_verso.lean"},
+ "position": {"line": 11, "character": 2}}
+[]
+{"textDocument":
+ {"uri": "file:///src/tests/interactive/test-cases/symbols_verso.lean"},
+ "position": {"line": 16, "character": 2}}
+[{"selectionRange":
+  {"start": {"line": 0, "character": 0}, "end": {"line": 0, "character": 12}},
+  "range":
+  {"start": {"line": 0, "character": 0}, "end": {"line": 11, "character": 0}},
```
ejgallego: The start here looks wrong. Let me look into it.

src/tests/interactive/test-cases/symbols_verso.lean
```diff
@@ -0,0 +1,18 @@
...
+set_option doc.verso true
+
+open Verso Genre Manual InlineLean
+
+#docs (Manual) docsSym "Docs heading" :=
+:::::::
+# Docs heading
+Text
+:::::::
+--^ textDocument/documentSymbol
```
ejgallego: Not sure this call is worth the effort, I'll keep it as a witness of the non-incrementality of documentSymbol

src/tests/interactive/test-cases/symbols_verso.lean.expected.out
```diff
@@ -0,0 +1,41 @@
...
+ {"selectionRange":
+  {"start": {"line": 17, "character": 0}, "end": {"line": 24, "character": 7}},
+  "range":
+  {"start": {"line": 17, "character": 0}, "end": {"line": 24, "character": 0}},
+  "name": "Docs heading two",
+  "kind": 15},
+ {"selectionRange":
+  {"start": {"line": 26, "character": 0}, "end": {"line": 33, "character": 7}},
+  "range":
+  {"start": {"line": 26, "character": 0}, "end": {"line": 33, "character": 0}},
```
ejgallego: There is a violation of the LSP spec here, selectionRange must be contained in range.

Working on a fix.

src/tests/interactive/test-cases/symbols_verso.lean.expected.out
ejgallego: Tried a fix but introduced a LSP violation (selectionRange not within range, see below)

src/tests/interactive/test-cases/symbols_verso.lean.expected.out
david-christiansen: Seems like we should fix that. Do you plan to do that in this PR?

david-christiansen:
> Heck yes. I wonder if there are reasons where/why it would be helpful to track the location of the `#doc` tag itself?

It's certainly useful for giving a good outline view and for populating the declaration ranges table.

src/verso/Verso/Doc/Concrete.lean
```diff
@@ -79,13 +79,13 @@ All-at-once elaboration of verso document syntax to syntax denoting a verso `Par
 elaboration of the `#docs` command and `#doc` term. The `#doc` command is incremental, and thus
 splits the logic in this function across multiple functions.
 -/
-private meta def elabDoc (genre: Term) (title: StrLit) (topLevelBlocks : Array Syntax) (endPos: String.Pos.Raw) : TermElabM Term := do
+private meta def elabDoc (rootStx : Syntax) (genre: Term) (title: StrLit) (topLevelBlocks : Array Syntax) (endPos: String.Pos.Raw) : TermElabM Term := do
```
david-christiansen: The meaning of `rootStx` isn't particularly apparent to me when just reading here. How about explaining in the docstring that it's the syntax of the complete `#doc ... =>` line?

david-christiansen: We might also want to call it `ref` for consistency with Lean internals if you think that makes sense

david-christiansen: Or could we drop the parameter and just call `getRef` here?

src/tests/interactive/test-cases/symbols_verso.lean.expected.out
ejgallego: Yes, the fix turned out to be more involved than I wanted, due to `header{}` being a synthetic token, which actually made the headings not carry location and not appear on the outline view.

ejgallego: This is indeed a serious violation, so I'll add a panic for this problem. In this case the origin of the issue was a little typo in the code `getPos` vs `getTailPos`.

src/verso/Verso/Doc/Concrete.lean
ejgallego: I'll find a better name, I agree.

Let me look into using `getRef`, as of now, in working branch, I'm doing a little bit of surgery but I agree it'd be better if we could use `getRef`.

ejgallego: Using getRef now.

ejgallego:
Thanks, I tried to address the remaining concerns in the latest push, see updated PR description.

IMHO this is ready for review and hopefully merge.

src/tests/interactive/test-cases/folding_verso_term_doc.lean
```diff
@@ -0,0 +1,39 @@
+import Verso
+import VersoManual
+
+set_option doc.verso true
+
+open Verso Genre Manual InlineLean
+open scoped Verso.Doc.Concrete
+
+-- NOTE: Verso currently recognizes markdown headers only when `#` starts in column 0.
```
david-christiansen:
```suggestion
-- NOTE: Verso requires headers to be well-nested and to have leading `#`s in column 0.
```

src/tests/interactive/test-cases/symbols_verso_term_doc.lean
```diff
@@ -0,0 +1,33 @@
+import Verso
+import VersoManual
+
+set_option doc.verso true
+
+open Verso Genre Manual InlineLean
+open scoped Verso.Doc.Concrete
+
+-- NOTE: Current Verso block parsing recognizes markdown headers only at column 0.
```
david-christiansen:
```suggestion
-- NOTE: Verso requires headers to be well-nested and to have leading `#`s in column 0.
```

src/tests/interactive/test-cases/symbols_verso_term_doc.lean
```diff
@@ -0,0 +1,33 @@
...
+-- Keep both term bodies flush-left; indented `#`/`##` lines would currently be parsed as paragraphs.
```
david-christiansen:
```suggestion
-- NOTE: Current Verso block parsing recognizes markdown headers only at column 0.
-- Keep both term bodies flush-left; indented `#`/`##` lines would be parsed as paragraphs.
```

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -167,8 +167,8 @@ public structure PartElabM.State where
   deferredBlocks : Array (Name × Term) := #[]
 deriving Inhabited
 
-public def PartElabM.State.init (title : Syntax) (expandedTitle : Option (String × Array (TSyntax `term)) := none) : PartElabM.State where
-  partContext := {titleSyntax := title, expandedTitle, metadata := none, blocks := #[], priorParts := #[], parents := #[]}
+public def PartElabM.State.init (rangeSyntax : Syntax) (selectionSyntax : Syntax := rangeSyntax) (expandedTitle : Option (String × Array (TSyntax `term)) := none) : PartElabM.State where
```
david-christiansen: Do we want `selectionSyntax` to have a default value here, or should callers actively pick?

src/verso/Verso/Doc/Concrete.lean
```diff
@@ -78,14 +78,19 @@ open PartElabM in
 All-at-once elaboration of verso document syntax to syntax denoting a verso `Part`. Implements
 elaboration of the `#docs` command and `#doc` term. The `#doc` command is incremental, and thus
 splits the logic in this function across multiple functions.
+
+The current elaborator reference anchors the root document part's TOC/LSP range. The title literal
+remains the root part's selection syntax, so document-symbol clients select the visible title rather
+than the entire document body.
```
david-christiansen: This reads more like a description of a change being made here than a description of what the code does once these changes are made.

I think we can just delete this paragraph, it's perhaps not so relevant to callers of this function but is more of an internals issue?

src/verso/Verso/Doc/Concrete.lean
```diff
@@ -78,14 +78,19 @@ open PartElabM in
...
 -/
 private meta def elabDoc (genre: Term) (title: StrLit) (topLevelBlocks : Array Syntax) (endPos: String.Pos.Raw) : TermElabM Term := do
+  let rootRangeStx ← getRef
   let env ← getEnv
   let titleParts ← stringToInlines title
   let titleString := inlinesToString env titleParts
   let ctx ← DocElabContext.fromGenreTerm genre
   let initDocState : DocElabM.State := { highlightDeduplicationTable := .some {} }
-  let initPartState : PartElabM.State := .init (.node .none nullKind titleParts)
+  let initPartState : PartElabM.State := .init rootRangeStx title.raw
```
david-christiansen, line 91:
```suggestion
  -- Use the complete document syntax for the range, and the title syntax for the selection.
  -- This means that the outline will contain all sub-parts, and clicking the root goes to the title.
  let initPartState : PartElabM.State := .init rootRangeStx title.raw
```

src/verso/Verso/Doc/Lsp.lean
```diff
@@ -230,10 +230,23 @@ where
...
+  stx.lspRange text <|> do
+    let start ← stx.getPos?
+    let «end» ← stx.getTailPos?
+    pure ⟨text.utf8PosToLspPos start, text.utf8PosToLspPos «end»⟩
+
 meta def rangeContains (outer inner : Lsp.Range) :=
   outer.start < inner.start && inner.«end» ≤ outer.«end» ||
   outer.start == inner.start && inner.«end» < outer.«end»
 
+-- In the LSP specification, `selectionRange` MUST be contained in `range`, otherwise UB
```
david-christiansen:
```suggestion
/--
In the LSP specification, `selectionRange` MUST be contained in `range`, otherwise UB.
Asserts this fact, panicking if it does not hold.
-/
```

src/verso/Verso/Doc/Lsp.lean
```diff
@@ -366,27 +382,18 @@ meta partial def handleSyms (_params : DocumentSymbolParams) (prev : RequestTask
...
               range := rng,
-              selectionRange,
+              selectionRange := assertSelectionRangeContained s!"PointOfInterest '{title}'" rng selectionRange,
               detail?,
               kind
             }
             -- mergeInto is needed here to keep the tree invariant
             syms := mergeInto sym syms
       pure syms
-    truncate (outer inner : Range) : Range :=
```
david-christiansen, line 382: Should we truncate even on panic when the assertion fails?

src/verso/Verso/Doc/Concrete.lean
david-christiansen: Perhaps also this?
```suggestion
  let initPartState : PartElabM.State := .init (← getRef) title.raw
```

david-christiansen:
I haven't read the source positions in the tests in detail, but I'll assume that you've got those checked and that the desired interactive behavior is present.

Modulo minor quibbles about default arguments and comments, this is great. Thanks!

src/verso/Verso/Doc/Elab/Basic.lean
```diff
@@ -44,9 +48,54 @@ public partial def FinishedPart.toSyntax [Monad m] [MonadQuotation m]
...
+Policy:
+
+- If the syntax has a recoverable range, preserve it unchanged.
+- If the range is unrecoverable, panic loudly. This indicates malformed/elided source provenance at TOC
+  construction time, and we want to fail early rather than emit invalid LSP data downstream.
+
+This keeps source-span invariants close to TOC construction (before LSP handlers), so later symbol
+assembly only consumes already-normalized syntax.
+-/
+private def normalizeTOCSyntax (title : String) (stx : Syntax) : Syntax :=
```
david-christiansen: This isn't doing any normalization, from what I can see.

ejgallego:
@david-christiansen I think this PR is in a good shape now. I had to dig a bit more into what was going on: I think the root of all issues was actually the Verso LSP code not using `lspRangeOfStx` API, which is the right one.

So the PR now :
- makes LSP code use lspRangeOfStx which does the right thing (TM) for Verso syntax.
- enforces that syntax has correct ranges before saving to the LSP facing database, else it panics.
- LSP code is simplied and now relies on the range metadata being correct, instead of trying to repair it
- we save both range and selection range for Verso constructions, we improve a few places here
- we fix a bug on setting the final position of a document, and a type mismatch when elaborating Verso in term syntax (added a test for this)

The end result is that outline and folding now work much better, tested interactively on the examples.

I feel good about merging this as-is, let me know what you think.

ejgallego:
Couple of extra comments / points:
- I did keep a few `let rootStx <- getRef` vs `(<- getRef)` later in the code as to guard better against ref manipulation in intermediate monadic calls. Likely I'm being too paranoid, but IMHO acceptable.
- The POI (Point of Interest) API still carries selectionRange as an option, but in LSP this is a required field. This was a bit more invasive and likely user-breaking so I decided not to attack that refactoring here.

david-christiansen (approved):
I think it looks great too!

PR #774, opened 2026-02-27
chore: bump to v4.29.0 rc2

david-christiansen:


david-christiansen:
I think this renders #769 unnecessary - I had a version of @kim-em's branch that I fixed it on.

PR #775, opened 2026-02-27
chore: enforce commit convention for PR title

david-christiansen:


ejgallego:
@david-christiansen is the commit convention of Verso documented in the repos? I can't find it.

david-christiansen:
It's the same as in Lean

david-christiansen:
We should write that down though!

PR #778, opened 2026-03-02
fix: delay processing of KaTeX

david-christiansen:
This PR delays the processing of KaTeX until files are actually being generated, rather than adding its assets to the traverse state. This makes traversal faster and allows clients to completely opt out of deserializing it.

### DIFF
--- a/src/verso-manual/VersoManual/Html/Features.lean
+++ b/src/verso-manual/VersoManual/Html/Features.lean
@@ -50,7 +50,6 @@ public structure HtmlFeatures where
   private hasSearch : Bool
 deriving Repr, Inhabited, BEq
 
-
 public instance : LawfulBEq HtmlFeature where
   rfl := by
     intro f; cases f <;> rfl


src/verso-manual/VersoManual/Html/Features.lean
```diff
@@ -66,85 +62,228 @@ public instance : LawfulHashable HtmlFeature where
...
 /--
 Converts a feature set into an array of features.
 -/
 public def toArray (features : HtmlFeatures) : Array HtmlFeature :=
-  features.features.toArray |>.qsortOrd
+  match features with
+  | ⟨true, true⟩ => #[.KaTeX, .search]
+  | ⟨true, false⟩ => #[.KaTeX]
+  | ⟨false, true⟩ => #[.search]
+  | ⟨false, false⟩ => #[]
```
robsimmons, line 160: Seems like this is immediately regrettable as soon as there's a third thing. How about

```suggestion
  #[ if features.hasKatex then #[.KaTeX] else #[],
     if features.hasSearch then #[.search] else #[], ].flatten
```

robsimmons: better, probably: keep the pattern match ⟨isKatex, isSearch⟩ so that there's a type error when a feature gets added

david-christiansen: That was the thought process! Though if we start getting more features, we will want some other implementation of this, akin to what you're suggesting, and we can easily mitigate this risk by proving an appropriate lemma.

PR #804, opened 2026-03-13
fix: avoid netlify rewriting

jcreedcmu:
Netlify appears to turn `Index/` into `index.html`, even when Pretty URLs is disabled. 
Consequently, 
https://verso.lean-lang.org/doc/latest/Index/#index
is broken, because it has the same content as
https://verso.lean-lang.org/doc/latest/index.html
which in particular means a `<base>` href of `"."`, which is very wrong.

Compare a similar fix that was done to the lean reference manual:
https://github.com/leanprover/reference-manual/commit/8acff2f6e6fbaf748dc6d9ce0faae54d5a0e1bca

### DIFF
--- a/doc/UsersGuide/Basic.lean
+++ b/doc/UsersGuide/Basic.lean
@@ -113,6 +113,7 @@ Mixing incompatible features results in an ordinary Lean type error.
 # Index
 %%%
 tag := "index"
+file := "the-index"
 number := false
 %%%
 


david-christiansen:
Turning off their so-called 'pretty URL' feature should address this, if I understand the docs correctly. Can you check that first?

david-christiansen:
It often needs a redeploy after fixing the setting

jcreedcmu:
> even when Pretty URLs is disabled.

david-christiansen:
Gah.

jcreedcmu:
I disabled it for both the verso website netlify porject (and redeployed the verso website project) and found it already disabled in the verso user manual netlify project.

david-christiansen (approved):

jcreedcmu:
Preview index looks good.

jcreedcmu:
> It often needs a redeploy after fixing the setting

I suspected the same, (since it's referred to as a 'postprocessing' step) and preemptively tried redeploying before filing this PR. But it didn't seem to help.

PR #825, opened 2026-04-08
chore: add !consistent PR-comment auto-fix for lean-toolchain drift

kim-em:
When the test-project `lean-toolchain` files drift from the root `lean-toolchain`, the consistency check now points the reader at a new comment-triggered workflow: commenting `!consistent` on the PR runs `cp lean-toolchain test-projects/*/lean-toolchain` and pushes the fix back to the PR branch, matching the existing `!prettier` pattern.

### DIFF
--- a/.github/workflows/consistent-pr-comment.yml
+++ b/.github/workflows/consistent-pr-comment.yml
@@ -70,8 +70,18 @@ jobs:
 
             - name: Sync test-project lean-toolchains
               run: |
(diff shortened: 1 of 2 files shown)


david-christiansen:
I think this might be nicer as a GH action? So we could run:

`!consistent`

and then it would fix all consistency issues in the obvious way and push a commit to the PR?

david-christiansen:
Thanks!

PR #835, opened 2026-04-12
Fix/relative urls

### LINKED ISSUE #834: BUG: Relative URLs broken by `<base>` tag in multi-page HTML output
## Summary

In multi-page HTML output, the use of the `<base>` tag currently shifts URL resolution in a way that breaks two distinct types of resources: user-authored relative images and system-injected KaTeX assets.

This issue aims to clarify the expected behavior for these links and document where the current multi-page output diverges from that expectation.

## Expected Behavior (Proposed Link Resolution Rules)

To ensure a predictable experience for Verso authors, links and assets should conceptually resolve as follows:

1. **User-authored relative paths:** If an author includes a relative link or image in a Markdown file (e.g., `diagram`), that path should resolve relative to the location of that specific page within the project structure.
2. **System-injected internal assets:** Internal Verso assets (like KaTeX scripts mapped to `/-verso-data/katex/...`) should reliably resolve to the generated site-root directory, regardless of how deeply nested the current page is.

## Reproduction

**Setup:**
- Verso `v4.29.0`, multi-page HTML output enabled.
- KaTeX feature enabled.
- A nested page created at the path: `Verification/Key-Theorems/`
- The Markdown content for that page includes a local image reference: `diagram`

## Actual Behavior 

Because the generator injects `<base href="./../../">` on the nested page, it alters the base URL for the entire document, breaking both rules defined above:

**1. Markdown Images (Overshoots local directory)**
The relative image `diagram` is rendered in the HTML as ``. 
Because of the `<base>` tag, the browser resolves this from the site root rather than the page directory.
* **Expected:** `{site-root}/Verification/Key-Theorems/graphs/foo.svg`
* **Actual:** `{site-root}/graphs/foo.svg` (404)

**2. KaTeX Assets (Overshoots site root)**
Absolute internal assets like `/-verso-data/katex/katex.js` are currently relativized by the compiler to `../../-verso-data/katex/katex.js`.
Because the `<base>` tag *already* navigates the browser back to the site root, the additional `../../` prefix causes the browser to overshoot the root entirely.
* **Expected:** `{site-root}/-verso-data/katex/katex.js`
* **Actual:** `{two-levels-above-site-root}/-verso-data/katex/katex.js` (404)

Browser console output shows:
`GET https://example.com/pages/-verso-data/katex/katex.js net::ERR_ABORTED 404`

## Proposed Fix

I have identified the root causes for both of these URL resolution issues in `Html.lean` and `VersoManual.lean` and opened a PR with a proposed fix here: https://github.com/leanprover/verso/pull/835

The PR adjusts the compiler to ensure user-authored relative paths are prefixed with their page path before the `<base>` tag shifts resolution, and prevents absolute asset URLs from generating redundant `../` sequences.

NicolasRouquette:
Fixes #834

## Problem

Two classes of URLs are broken in multi-page HTML output:

1. **Relative image URLs** (from Markdown `alt`) resolve from the site root instead of the page directory, because `<base href="./../../">` shifts resolution.
2. **Absolute internal asset URLs** (`/-verso-data/katex/...`) are converted by `path.relativize` to `../../-verso-data/...`, which overshoots the root when combined with the `<base>` tag.

### `VersoManual/Html.lean`

- **`rwAttr`**: Replace `path.relativize` with simple `/`-stripping (`(attr.snd.drop 1).toString`). Since `<base>` already points to the site root, root-relative paths resolve correctly without `../` sequences.
- **`rwTag`**: For `` tags with relative `src` paths, prefix with the page path before calling `rwAttr`. This distinguishes user-authored relative URLs (need page-path prefix) from absolute URLs (just need `/` stripped). Adds `isRelativeContentUrl` helper.

### `VersoManual.lean`

- **`relativizeLinks`**: Now accepts a `Path` parameter and passes it to `Html.relativize`, so the page path is available for image URL adjustment.
- Updated all three call sites: `emitFindHtml` (`#["find"]`), single-page emitter, and `emitPart`.

## Result

| URL type | Before | After |
|----------|--------|-------|
| Image (`graphs/foo.svg` on page `Verification/Key-Theorems/`) | `src="graphs/foo.svg"` → resolves to `{root}/graphs/foo.svg` (404) | `src="Verification/Key-Theorems/graphs/foo.svg"` → correct |
| KaTeX asset | `src="../../-verso-data/katex/katex.js"` → overshoots root (404) | `src="-verso-data/katex/katex.js"` → correct |
| Internal links | `href="../../Overview/"` → correct | unchanged |
| CSS/JS (`book.css`, `-verso-search/*`) | correct | unchanged |


robsimmons:
The description of #834 is a bit unfocused right now — however, rather than asking you to tighten that up, I think this primarily exposes that we need to clearly document the *expected* behavior of links in the Verso manual, and then we can more clearly pinpoint whether you're describing behavior that differs from the expected behavior. We're fine keeping this open in the meantime.

NicolasRouquette:
Thanks for the feedback and context, @robsimmons! That makes complete sense. I completely agree that establishing the intended "rules" for how links should resolve in Verso is the necessary first step here.

Even though you didn't ask me to, I went ahead and updated the issue description to be much more focused on the user-facing side of things rather than the code implementation. I've drafted a proposed set of "Expected Behaviors" for how user-authored Markdown links and system-injected assets should resolve.

Could you take a look at the updated description and let me know if those proposed rules align with the intended design for the Verso manual? If we can agree on those rules, it should make it much easier to evaluate whether the current behavior (and my proposed PR) hits the mark. Happy to adjust the proposal if the project has a different vision for link resolution!

robsimmons:
The rewritten issue is a better description of the problem at hand — this is a thing that we want to handle consistently across Verso genres but this is not the approach that we want to take, so I'm going to close this PR for now.

PR #846, opened 2026-04-23
feat: for HTML output, add Table of Contents resize handle

### LINKED ISSUE #845: Need resize handle for the table of contents for HTML output
Currently the navigation pane for the tables of contents is a fixed width. For some documents, e.g. ones that follow a file tree hierarchy, the fixed width is too narrow. It would be helpful if there were a resize handle for the navigation pane.

richardlford:
Adds a drag handle on the right edge of the ToC in HTML manual output so readers can set its width, with the choice persisted in localStorage. Pointer and keyboard resizing are both supported, and the handle is exposed as an ARIA separator.

The layout is entirely CSS-driven. The JavaScript (static-web/toc-resize.js) only records the user's preferred width in the --verso-toc-user-width custom property and in localStorage.

There are two scripts on purpose:

  - toc-resize-preload.js runs inline in the <head> and applies the saved width before first paint, so returning readers don't see the default width flash to their saved width.

  - toc-resize.js is deferred, re-applies the width defensively, and wires up the interactive resizing.

Both must agree on the localStorage key, since the preload primes the value the main script reads. A CI step enforces that the keys match.

The content area's width depends on the ToC width, not just the viewport. This means that some media queries became container queries, and some thresholds were adjusted (for an equivalent result of the old thresholds).

Closes #845


richardlford:
It has been about 5 weeks since I opened this PR. I'm guessing it has
not been reviewed because the code was written by Claude Code.
Or that a detailed description of the changes was not given.

So I have now edited the PR description to include 
a detailed description of the changes and the reasons for them.
I believe all of the changes are needed and have a specific purpose.

richardlford:
@david-christiansen,  is there anything else you would like me to do before this is reviewed?

richardlford:
I have changed the PR title so the subject does not start with a capital letter.

david-christiansen:
Sorry for the slow response here.

This is a good feature to have, and thank you for implementing it! There's a few things I think it should do differently (e.g. keyboard accessibility/aria, putting JS code in separate files and adding to tsc), but I'll just do those and merge because you've waited so long.

Also, our system makes the PR description into the commit message, so I'll move the current one into a PR comment for posterity and then adapt the description prior to merge.

Thanks again!

david-christiansen:
Here's @richardlford's original PR description, for posterity. I'm editing the PR description before merging because it ends up as the commit message.

-----------

Fixes #845

This change adds a resize handle for the navigation pane that holds the table of contents.

The code itself was generated using Claude Code.

Since the code was written by AI, to assist reviewers and other readers, I'll
now describe the changes in detail. The description has some questions
in it that reviewers should try to answer.

In the `src/verso-manual/VersoManual/Html.lean` file
there are the following changes:

The `page` function (in namespace `Verso.Genre.Manual.Html`) is a function which
composes the HTML for a page of a manual document. This PR makes two additions:

```
<div class="toc-resize-handle" aria-hidden="true"/>
```

The above inserts an element that can be used to resize the
width of the table of content navigation area.

- It is given a class so that CSS can style it.

- aria-hidden="true" tells screen readers and other assistive
  technologies to completely ignore this element — it will not be
  announced, focusable, or part of the accessibility tree. The reason
  it's appropriate here: the resize handle is a purely visual/pointer
  interaction with no meaningful semantic content. A screen reader
  user cannot drag it, and there's nothing useful to announce about
  it. Without aria-hidden, a screen reader might land on it and read
  something unhelpful like "unlabeled div" or "group." Hiding it keeps
  the accessibility experience clean.


```
<script>{{Html.text false tocResizeJs}}</script>
```

This above has the effect of inserting the Javascript code
defined by the tocResizeJs string defined above
(and to be described in detail below). The `false`
means do not escape things like '&' to '&amp;amp;'.
From what I've read on the web, putting the script
at the end (i.e., in the &lt;body&gt;) makes sure all the elements it may need
to reference have been created.


Above the modified `page` function is a definition of a string
named `tocResizeJs`. This is a raw Lean string that holds
Javascript code to handle the resizing operation. It is
defined in a separate string to avoid cluttering the `page`
function.

The string defines a Javascript function (a closure)
which does the following:

- Gets the elements for the resize handle and the toc. If either of
these does not exist it just returns.

- Defines constants for the minimum and maximum toc widths
- Defines a key in the browser local storage to save the
  toc width.
- Looks up the saved width, if any, from local storage.
  If a saved value was found, use it to set the "--verso-toc-width"
  property.
- Initialize these variables to prepare for resizing:
  - dragging: tell if we are currently dragging with mouse
  - startX: Save the X coordinate when dragging started.
  - startWidth: The width of the toc when dragging started.
- Add an event listener on the toc resize handle for `mousedown`
  which does the following:
  - If the browser window is less than 700px, return and do
    nothing. I'm not sure about this. Why this number?
    I do notice that when I make the browser window narrower
    that the navigation pane disappears entirely
    (independent of this PR).
  - Set `dragging` to `true` and record `startX` and `startWidth`.
  - Adds "dragging" to the class list for the resize handle so
    CSS can style it differently.
  - `document.body.style.userSelect = 'none';` prevents users
    from accidental text highlighting while dragging.
  - `document.body.style.cursor = 'col-resize';` changes the
    mouse icon to a horizontal resize icon.
  - `e.preventDefault();` tells the browser to skip its usual
    mousedown action.
- Add an event listener on the toc resize handle for `mousemove` events,
  which does the following:
  - If not dragging, return and do nothing.
  - Compute a new toc width based on the current x value,
    clamped using the minimum and maximum values.
  - Set the "--verso-toc-width" CSS property with the new value.
- Add an event listener on the toc resize handle for `mouseup` events,
  which does the following:
  - If not dragging, return
  - Set dragging to false
  - Remove "dragging" from the handles class list
  - `document.body.style.userSelect = '';` no longer inhibits selections.
  - `document.body.style.cursor = '';` returns to normal mouse icon
  - Save the new width in browser local storage

In the `src/verso-manual/VersoManual/Html/Style.lean` file the following
additions were added:

```
.toc-resize-handle {
    position: absolute;
    top: 0;
    right: 0;
    width: 6px;
    height: 100%;
    cursor: col-resize;
    z-index: 12;
}
```

This has the following meaning for the resize element:

- .toc-resize-handle: A class selector targeting the specific HTML
  element meant to act as the adjustment handle.

- position: absolute;: Removes the element from normal document
  flow. It places it relative to its closest parent container that has
  a position: relative, absolute, or fixed styling.
  In this case, its parent will be the navigation pane

- top: 0; right: 0;: Anchors the handle precisely to the top-right
  corner of that parent container.

- width: 6px;: Gives the handle a narrow interactive hit-box of 6
  pixels.

- height: 100%;: Stretches the handle vertically to span the full height of its parent container.

- cursor: col-resize;: Changes the mouse pointer to a horizontal
  two-headed arrow (\(\leftrightarrow \)) when hovering. This visually
  informs the user they can drag it horizontally.

- z-index: 12;: Layers the handle on top of surrounding content. This
  ensures text or images behind it do not block the user from clicking
  and dragging it.

Note that the above does not yet make the toc handle appear.

```
.toc-resize-handle::after {
    content: '';
    position: absolute;
    top: 0;
    left: 3px;
    width: 1px;
    height: 100%;
    background: var(--verso-toc-border-color, #d0d0d0);
    transition: background 0.15s;
}

.toc-resize-handle:hover::after,
.toc-resize-handle.dragging::after {
    background: var(--verso-toc-resize-handle-color, #888);
}
```

The ::after pseudo-element is a technique for separating the
clickable/draggable hit area from the visible indicator line.

The .toc-resize-handle div itself is 6px wide — wide enough to be easy
to grab with a mouse — but it has no visible background, so it's
invisible.

The ::after pseudo-element renders the actual 1px visual line,
centered within the 6px hit area (left: 3px). This is the subtle
border you see between the TOC and the main content.

The benefit of this separation: the visual line stays thin (1px) and
unobtrusive, while the grabbable target is 6× wider, making it much
easier to click. If the 1px line were the hit target, it would be
frustratingly hard to grab precisely.

The ::after rule for hover and dragging then changes the line's color
from light gray (#d0d0d0) to a darker gray (#888) when the user hovers
or is actively dragging, giving visual feedback that the handle is
interactive. The transition: background 0.15s  makes that
color change animate smoothly rather than snap.

```
@media screen and (max-width: 700px) {
    .toc-resize-handle {
        display: none;
    }
}
```

This CSS rule hides the resize handle on mobile screens and small
windows that are 700 pixels wide or smaller.

That completes the description of the changes for this PR.

david-christiansen (approved):

david-christiansen:
@richardlford can you please confirm that this still does what you needed it to do before I merge?

Thanks again for your patience and friendliness!

richardlford:
> @richardlford can you please confirm that this still does what you needed it to do before I merge?
> 
> Thanks again for your patience and friendliness!

Yes. It works fine. Thank you for your improvements. I'm not an HTML expert and will learn from your changes.

PR #847, opened 2026-04-23
feat: search page

david-christiansen:
Adds a full-page search, accessed by pressing enter in the search field. This page shows more results, provides more context, and has more filtering features.


robsimmons (approved):
There's a couple of places (especially the search-page.js state management object) where I'm tempted to suggest using classes — but the real issue insofar as there is one is that this is pushing up against the constraints of our minimalistic dependency usage — the `renderToken` infrastructure is a placeholder for react useEffect style cancellation hooks, for example. But I don't want to delay this getting in any further, and it works for what its trying to do and seems in line with the code/tech we already have.

PR #854, opened 2026-05-10
perf: avoid re-exporting imported doc expanders

ejgallego:
Verso Blueprint exposed very large `.olean` files. For example, a file that imports all test blueprints produced an `.olean` above 10 GiB.

The cause was that doc expander and signature environment extension state were re-exported. Importer modules therefore serialized transitive expander entries again, which can compound across import chains.

This patch introduces `LocalPersistentEnvExtension`, a small Verso-side helper for persistent extensions that keep complete imported lookup state alongside a separate local export delta. It then uses the helper to store doc expander and signature extension state as separate local and complete maps.

Only local entries are exported, while lookup still uses the complete imported state. This prevents importer modules from reserializing every transitive expander entry.

The patch adds transitive-import regressions covering role, code block, directive, and block-command expanders, plus a direct regression for the local-persistent-extension failure mode.

## Measurements

For blueprints, the size reduction is substantial, on the order of 100x or more.

Additional benchmarks:

Reference manual at `8ae381a3`, Lean `v4.30.0-rc2`, command `lake --no-cache build Manual`, baseline Verso `ba73c230`:

| Build | Elapsed | Max RSS |
| --- | ---: | ---: |
| baseline run 1 | 157.13s | 2,518,464 KB |
| baseline run 2 | 155.61s | 2,525,236 KB |
| patched | 144.25s | 2,537,208 KB |

Compared reference-manual artifacts:
- total artifacts: `637,348,589 B -> 599,759,959 B`, `-37,588,630 B` / about `-35.8 MiB`
- `.olean` total: `546,283,272 B -> 508,649,832 B`, `-37,633,440 B`

Synthetic importer test, `Manual.PerfImportPair` importing `Manual.Terms` and `Manual.Tactics.Reference`:
- Lake-built `.olean`: `375,456 B -> 17,072 B`, `-95.5%`
- direct Lean `.olean`: `375,392 B -> 17,008 B`
- direct compile time stayed flat: `1.66s -> 1.65s`

UsersGuide comparison:
- elapsed time effectively flat: `56.59s -> 56.84s`
- compared `.olean` artifacts dropped by about `432 KiB`

### DIFF
--- a/src/tests/Tests/DocElabExtensions/Define.lean
+++ b/src/tests/Tests/DocElabExtensions/Define.lean
@@ -1,0 +1,40 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Emilio Jesus Gallego Arias
+-/
+import Tests.DocElabExtensions.LocalExtension
+import VersoManual
+
+namespace Verso.Tests.DocElabExtensions
+
+/-!
+This module defines the local entries used by the transitive-import regression.
+It registers one test-only local-persistent entry and one custom expander for each document
+extension kind. `Middle` imports this module without adding entries, and `Use` imports `Middle`
+to check that transitive lookup works without re-exporting `Middle`'s imported entries.
+-/
+
+open Verso Genre Manual
+open Verso.Doc.Elab
+open Lean
+
+#register_inherited_test_local_entry
+
+@[role]
+def inheritedRole : RoleExpanderOf Unit
+  | (), _contents => ``(Doc.Inline.text "inherited role")
+
+@[code_block]
+def inheritedCode : CodeBlockExpanderOf Unit
+  | (), str => ``(Doc.Block.code $(quote str.getString))
+
+@[directive]
+def inheritedDirective : DirectiveExpanderOf Unit
+  | (), blocks => do
+    let blocks ← blocks.mapM elabBlock
+    ``(Doc.Block.concat #[$blocks,*])
+
+@[block_command]
+def inheritedCommand : BlockCommandOf Unit
+  | () => ``(Doc.Block.para #[Doc.Inline.text "inherited command"])
--- a/src/tests/Tests/DocElabExtensions/LocalExtension.lean
+++ b/src/tests/Tests/DocElabExtensions/LocalExtension.lean
@@ -1,0 +1,51 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Emilio Jesus Gallego Arias
+-/
+import Lean.Elab.Command
+import Verso.EnvExtension
+
+namespace Verso.Tests.DocElabExtensions
+
+/-!
+This module defines a small test-only `LocalPersistentEnvExtension`.
+`Define` adds one local entry, `Middle` imports `Define` without adding entries, and `Use` checks
+that lookup in the final importer sees the entry while `Middle`'s serialized entries for this
+extension are empty.
+-/
+
+open Lean Elab Command
+
+private def addTestEntry (entries : Lean.NameMap (Array Name)) (key value : Name) :
+    Lean.NameMap (Array Name) :=
+  entries.insert key <| (entries.find? key |>.getD #[]).push value
+
+private abbrev TestLocalExtension :=
+  LocalPersistentEnvExtension (Name × Array Name) (Name × Name) (Lean.NameMap (Array Name))
+
+initialize testLocalExt : TestLocalExtension ←
+  LocalPersistentEnvExtension.register {
+    name := `Verso.Tests.DocElabExtensions.testLocalExt
+    mkInitialState := pure {}
+    addImportedEntryFn
+      | entries, (key, values) =>
+        values.foldl (init := entries) fun entries value =>
+          addTestEntry entries key value
+    addEntryFn
+      | entries, (key, value) =>
+        addTestEntry entries key value
+    exportEntriesFn _ entries :=
+      .uniform entries.toArray
+  }
+
+def testLocalExtensionEntries (env : Environment) (key : Name) : Array Name :=
+  testLocalExt.getState env |>.find? key |>.getD #[]
+
+def testLocalExtensionModuleEntries (env : Environment) (moduleIdx : ModuleIdx) :
+    Array (Name × Array Name) :=
+  testLocalExt.getModuleEntries env moduleIdx
+
+elab "#register_inherited_test_local_entry" : command => do
+  modifyEnv fun env =>
+    testLocalExt.addEntry env (`inheritedTestLocalExtension, `inheritedTestEntry)
--- a/src/tests/Tests/DocElabExtensions/Middle.lean
+++ b/src/tests/Tests/DocElabExtensions/Middle.lean
@@ -1,0 +1,16 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Emilio Jesus Gallego Arias
+-/
+import Tests.DocElabExtensions.Define
+
+namespace Verso.Tests.DocElabExtensions
+
+/-!
+This module is the middle import in the transitive-import regression.
+It imports the local entries from `Define` but intentionally adds no new entries itself, so `Use`
+can check that this module does not re-export inherited local-persistent entries.
+-/
+
+def importedThroughMiddle : Bool := true
--- a/src/tests/Tests/DocElabExtensions/Use.lean
+++ b/src/tests/Tests/DocElabExtensions/Use.lean
@@ -1,0 +1,50 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Emilio Jesus Gallego Arias
+-/
+import Tests.DocElabExtensions.Middle
+
+namespace Verso.Tests.DocElabExtensions
+
+/-!
+This module consumes the entries through `Middle`.
+The command guard checks the test-only local-persistent extension directly: the final lookup state
+contains the entry from `Define`, but `Middle` exported no entries for the extension. The `#docs`
+block separately checks the real doc expander registries for role, code block, directive, and block
+command expanders through the same transitive import.
+-/
+
+open Verso Genre Manual
+open Lean Elab Command
+
+elab "#guard_middle_exports_no_local_persistent_entries" : command => do
+  let env ← getEnv
+  let some middleIdx := env.getModuleIdxFor? ``importedThroughMiddle
+    | throwError "could not find the module for importedThroughMiddle"
+  let entries := testLocalExtensionEntries env `inheritedTestLocalExtension
+  unless entries.contains `inheritedTestEntry do
+    throwError "expected lookup state to include the entry imported through the middle module"
+  let middleEntries := testLocalExtensionModuleEntries env middleIdx
+  unless middleEntries.isEmpty do
+    throwError "middle module should export no inherited local-persistent entries, but exported {middleEntries.size}"
+
+#guard_middle_exports_no_local_persistent_entries
+
+#docs (Manual) inheritedDocElabExtensions "Inherited Doc Elab Extensions" :=
+:::::::
+This custom {inheritedRole}[] is provided through a transitive import.
+
+```inheritedCode
+code block body
+```
+
+:::inheritedDirective
+Directive body
+:::
+
+{inheritedCommand}
+:::::::
+
+#guard importedThroughMiddle
+#guard inheritedDocElabExtensions.toPart.content.size == 4
--- a/src/verso/Verso/Doc/Elab/Monad.lean
+++ b/src/verso/Verso/Doc/Elab/Monad.lean
@@ -14,6 +14,7 @@ import Lean.DocString
 
 import SubVerso.Highlighting
 import Verso.Doc
+import Verso.EnvExtension
 public import Verso.Doc.ArgParse
 public import Verso.Doc.Elab.InlineString
 meta import Verso.Doc.Elab.InlineString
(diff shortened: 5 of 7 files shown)


ejgallego:
More generally, I wonder if it would be possible to improve Lean's upstream API to capture this use case in a more principled way.

david-christiansen:
Nice catch!

david-christiansen (approved):

ejgallego:
@david-christiansen , I've done some further refactoring of the code, adding a new helper module.

Let me know if the new version looks OK, TIA.

src/tests/Tests/DocElabExtensions/Define.lean
```diff
@@ -0,0 +1,34 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
```
david-christiansen, line 5: This is missing the Author tag

src/tests/Tests/DocElabExtensions/Define.lean
```diff
@@ -0,0 +1,34 @@
...
+def inheritedDirective : DirectiveExpanderOf Unit
+  | (), blocks => do
+    let blocks ← blocks.mapM elabBlock
+    ``(Doc.Block.concat #[$blocks,*])
+
+@[block_command]
+def inheritedCommand : BlockCommandOf Unit
+  | () => ``(Doc.Block.para #[Doc.Inline.text "inherited command"])
+
+end Verso.Tests.DocElabExtensions
```
david-christiansen:
```suggestion
```

src/tests/Tests/DocElabExtensions/Define.lean
```diff
@@ -0,0 +1,34 @@
...
+import Tests.DocElabExtensions.LocalExtension
+import VersoManual
+
+namespace Verso.Tests.DocElabExtensions
+
+open Verso Genre Manual
+open Verso.Doc.Elab
+open Lean
+
+#register_inherited_test_local_entry
```
david-christiansen, line 22: This file could really use a comment explaining what it tests, and how it tests it. I don't look forward to reverse-engineering it if it ever fails.

src/tests/Tests/DocElabExtensions/LocalExtension.lean
```diff
@@ -0,0 +1,45 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
```
david-christiansen, line 5: Missing Author tag

src/tests/Tests/DocElabExtensions/LocalExtension.lean
```diff
@@ -0,0 +1,45 @@
...
+
+def testLocalExtensionModuleEntries (env : Environment) (moduleIdx : ModuleIdx) :
+    Array (Name × Array Name) :=
+  testLocalExt.getModuleEntries env moduleIdx
+
+elab "#register_inherited_test_local_entry" : command => do
+  modifyEnv fun env =>
+    testLocalExt.addEntry env (`inheritedTestLocalExtension, `inheritedTestEntry)
+
+end Verso.Tests.DocElabExtensions
```
david-christiansen:
```suggestion
```

src/tests/Tests/DocElabExtensions/LocalExtension.lean
```diff
@@ -0,0 +1,45 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+import Lean.Elab.Command
+import Verso.EnvExtension
+
+namespace Verso.Tests.DocElabExtensions
+
+open Lean Elab Command
+
+private def addTestEntry (entries : Lean.NameMap (Array Name)) (key value : Name) :
```
david-christiansen, line 20: This file is also missing a description of what it tests, how it tests it, and its relationship to other test files.

src/tests/Tests/DocElabExtensions/Middle.lean
```diff
@@ -0,0 +1,11 @@
...
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+import Tests.DocElabExtensions.Define
+
+namespace Verso.Tests.DocElabExtensions
+
+def importedThroughMiddle : Bool := true
+
+end Verso.Tests.DocElabExtensions
```
david-christiansen:
```suggestion
```

I suppose there should be CI looking for this. The AIs keep wanting to insert them and it's hard to fight back manually.

src/tests/Tests/DocElabExtensions/Middle.lean
```diff
@@ -0,0 +1,11 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
```
david-christiansen, line 5: Missing author tag

src/tests/Tests/DocElabExtensions/Middle.lean
```diff
@@ -0,0 +1,11 @@
...
+import Tests.DocElabExtensions.Define
+
+namespace Verso.Tests.DocElabExtensions
+
+def importedThroughMiddle : Bool := true
```
david-christiansen: This file should also describe its relationship to the wide testing system (what it tests, how, and relationship to other test files)

src/tests/Tests/DocElabExtensions/Use.lean
```diff
@@ -0,0 +1,43 @@
+/-
+Copyright (c) 2026 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
```
david-christiansen, line 5: Author tag

src/tests/Tests/DocElabExtensions/Use.lean
```diff
@@ -0,0 +1,43 @@
...
+Directive body
+:::
+
+{inheritedCommand}
+:::::::
+
+#guard importedThroughMiddle
+#guard inheritedDocElabExtensions.toPart.content.size == 4
+
+end Verso.Tests.DocElabExtensions
```
david-christiansen:
```suggestion
```

src/tests/Tests/DocElabExtensions/Use.lean
```diff
@@ -0,0 +1,43 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
+import Tests.DocElabExtensions.Middle
+
+namespace Verso.Tests.DocElabExtensions
+
+open Verso Genre Manual
+open Lean Elab Command
+
+elab "#guard_middle_exports_no_local_persistent_entries" : command => do
```
david-christiansen, line 21: Please describe this test and how it tests what it tests so we don't have to reverse-engineer it.

src/verso/Verso/Doc/Elab/Monad.lean
```diff
@@ -549,18 +550,19 @@ unsafe def blockExpandersForUnsafe (x : Name) : DocElabM (Array BlockExpander) :
...
+        entries.insert x y
     addEntryFn
-      | xs, (x, y) =>
-        xs.insert x y
-    exportEntriesFn xs :=
-      xs.toArray
+      | entries, (x, y) =>
+        entries.insert x y
+    exportEntriesFn _ entries :=
+      .uniform entries.toArray
```
david-christiansen, line 578: Normally we sort these things in the array, and then do direct array lookups via binary search. This approach is going to add an overhead to all downstream modules, causing them to initialize the local data structure with addImportedEntryFn. Is there a reason for initializing each time? Is it in response to a particular slowdown? If so, we should have a comment to that effect (I could imagine that this is what's needed; once you're writing Verso, you're typically not importing into many downstream non-Verso-doc libraries, and we may not have a module index in order to make the binary search approach fast)

src/verso/Verso/EnvExtension.lean
```diff
@@ -0,0 +1,80 @@
...
+
+public def getModuleEntries {α β σ : Type} [Inhabited σ]
+    (ext : LocalPersistentEnvExtension α β σ) (env : Environment) (moduleIdx : ModuleIdx)
+    (level := OLeanLevel.exported) : Array α :=
+  match ext with
+  | ⟨rawExt⟩ => PersistentEnvExtension.getModuleEntries (level := level) rawExt env moduleIdx
+
+end LocalPersistentEnvExtension
+
+end Verso
```
david-christiansen:
```suggestion
```

src/verso/Verso.lean
```diff
@@ -19,6 +19,7 @@ public import Verso.Doc.Html
 public import Verso.Doc.Lsp
 public import Verso.Doc.Suggestion
 public import Verso.Doc.TeX
+public import Verso.EnvExtension
```
david-christiansen: Isn't this more part of the implementation than the interface? I think it's OK to not public-import it here, but just use it in Monad.lean

david-christiansen:
One substantial question, and lots of nitpicks.

Also, before a merge we need to make sure that the PR description makes sense for the Git history, because it becomes the commit message. It's good to have things like measurements but I think we should purge weird AI language that doesn't make sense.

src/tests/Tests/DocElabExtensions/Define.lean
ejgallego: Fixed, thanks!

src/tests/Tests/DocElabExtensions/Define.lean
ejgallego: Added a comment, let me know if that's clear enough or if I should expand it.

src/tests/Tests/DocElabExtensions/Define.lean
ejgallego: Fixed. I think the code formater would be the right place to handle style issues such as this one.

src/verso/Verso/Doc/Elab/Monad.lean
ejgallego: Good point.

The reason is that the patch preserves the existing eager lookup-state initialization.

For doc expanders, repeated keyed lookup during elaboration may be the hot path, while I'd expect import overhead to be low due to few entries really being exported in the common case.

What do you think? Happy to change the approach of add a code comment as you suggested.

ejgallego:
> Also, before a merge we need to make sure that the PR description makes sense for the Git history, because it becomes the commit message. It's good to have things like measurements but I think we should purge weird AI language that doesn't make sense.

Thanks for the careful review David. Small issues should be addressed.

Regarding the import approach, see the comment above, I think the way it is makes sense, but happy to update.

I've tried my best improving the language in the description, let me know if that looks better now.

src/tests/Tests/DocElabExtensions/Define.lean
```diff
@@ -0,0 +1,40 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: Emilio Jesus Gallego Arias
+-/
+import Tests.DocElabExtensions.LocalExtension
+import VersoManual
+
+namespace Verso.Tests.DocElabExtensions
+
+/-!
+This module defines the local entries used by the transitive-import regression.
```
david-christiansen, line 12: Here's how I'd write it:

```suggestion
This module defines the local entries used by the transitive-import regression test.
Previously, Verso duplicated the expanders table in each module's persistent environment
extension state, leading to a quadratic size blowup; this test ensures that each module is
saving only its new entries.

```

This way, it's clear what's being tested in a year or two when we've got 800 more PRs behind us.

src/verso/Verso/Doc/Elab/Monad.lean
david-christiansen: I think this makes sense. A comment to that effect would be nice, just because it's different from the most common pattern; another reason to do it this way is that we don't require expanders to be registered in the same module as their associated identifier, so we'd have to check _all_ imported arrays.

src/verso/Verso/EnvExtension.lean
```diff
@@ -0,0 +1,82 @@
...
+
+This is the same persistence discipline as `SimplePersistentEnvExtension`, staged for extension
+shapes where the serialized entry type and the added entry type differ. The complete lookup state and
+local export delta intentionally have the same type here. Extensions that need different state types,
+async support, or replay support should use `PersistentEnvExtension` directly.
+
+Imported entries are eagerly folded into the complete lookup state when the module is imported. This
+matches the existing Verso doc-expander lookup pattern, where document elaboration repeatedly queries
+the complete map by key. Extensions whose import-time initialization cost matters more than direct
+lookup should use `PersistentEnvExtension` directly and keep a custom serialized representation.
```
david-christiansen, line 29: "Existing" seems misleading here. Here's how I'd phrase it:

```suggestion
Imported entries are eagerly folded into the complete lookup state when the module is imported. This
is because document elaboration repeatedly queries the complete map by key. Extensions whose
import-time initialization cost matters more than direct lookup should use `PersistentEnvExtension`
directly and keep a custom serialized representation.
```

src/verso/Verso/EnvExtension.lean
```diff
@@ -0,0 +1,82 @@
...
+-/
+public structure LocalPersistentEnvExtension (α β σ : Type) where
+  private ext : PersistentEnvExtension α β (σ × σ)
+
+namespace LocalPersistentEnvExtension
+
+public instance {α β σ : Type} [Inhabited σ] :
+    Nonempty (LocalPersistentEnvExtension α β σ) :=
```
david-christiansen: Can't this be an `Inhabited` instance?

david-christiansen: If it doesn't work, then a comment about why would be useful. But not a big deal.

ejgallego: I've checked, and `initialize` takes either Inhabited or Nonempty.

Using `deriving Nonempty` doesn't work due to Nonempty deriving procedure being weaker than Inhabited. On the other hand, using `deriving Inhabited` does work, so I'll go that route and use:

```lean4
public structure LocalPersistentEnvExtension (α β σ : Type) where
  private ext : PersistentEnvExtension α β (σ × σ)
deriving Inhabited
```

src/verso/Verso/Doc/Elab/Monad.lean
ejgallego: Comment added:
> -- We eagerly rebuild a `NameMap` from imported entries instead of keeping the
-- exported arrays sorted and doing binary search at lookup time, which is the
-- more common pattern for local persistent extensions.
--
-- This preserves the existing eager lookup-state initialization: doc expander
-- signatures are looked up by name repeatedly during elaboration, so the hot path
-- should be a single map lookup. The import-time cost should be small in the
-- expected case, since few signatures are exported by typical modules.
--
-- Another reason not to rely on per-module sorted arrays is that expanders are
-- not required to be registered in the same module as their associated
-- identifier, so lookup would otherwise need to search all imported arrays unless
-- we built an index anyway.

src/verso/Verso/EnvExtension.lean
ejgallego: It seems that in general Inhabited is preferred, so going this way.

PR #859, opened 2026-05-22
fix: data handlers for Verso docstring extensions in literate mode

david-christiansen:
Now, we warn instead of error on missing handlers. This PR also adds missing builtin handlers for code blocks and fixes a bug where the lookup for user handlers occurred in the wrong table. In particular, `getBlockToLiterateUnsafe` was looking in the inline handler table instead of the block handler table.


test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+{tactic}`rfl` is a tactic; {conv}`rfl` is a conv tactic.
+
+{attr}`@[simp]` is an attribute.
+
+{syntaxCat}`term` is a syntax category, and {syntax term}`1 + 2` is syntax in it.
+
+{kw (of := Lean.«command__Unif_hint____Where_|_-⊢__»)}`unif_hint` is a keyword atom.
+
+{option}`maxHeartbeats` is an option name; {option}`set_option maxHeartbeats 1000` is a `set_option`.
```
robsimmons, line 38: This shows up as `set_option [anonymous]`

Image

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+{lit}`anything goes here` is a literal code element.
+
+{manual section "some-section"}`a manual link` references a section.
+
+The {assert}`Nat.zero = Nat.zero` defeq assertion.
```
robsimmons, line 44: This isn't obviouly a problem but it seemed unexpected to me that I didn't get any hovers for `Nat.zero`.

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+The {lit}`assert'` role also runs an equality check and can be used in bootstrapping.
```
robsimmons: This doesn't actually demonstrate the use of this role, I wasn't able to figure out how to use it, and while it appears to be defined in the lean core this is the only reference to it in the verso codebase.

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+```lean (name := nb)
+#eval 1 + 1
+```
+
+```output nb
+2
+```
+-/
+def codeBlockDemos : Nat := 0
```
robsimmons, line 69: The output blue-squiggly is showing up in the output in the wrong place, atop the `def`

Image

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+{lit}`anything goes here` is a literal code element.
+
+{manual section "some-section"}`a manual link` references a section.
+
+The {assert}`Nat.zero = Nat.zero` defeq assertion.
+
+The {lit}`assert'` role also runs an equality check and can be used in bootstrapping.
+
+Given {given}`n : Nat`, the function operates on {name}`n`. Given {givenInstance}`Add Nat`, addition is available.
```
robsimmons, line 46: This actually tests the behavior:

```suggestion
Given {given}`n : Nat`, the function operates on {name}`n`.

Given {givenInstance}`Add String`, suddenly {lean}`"A" + "B"` is possible, where it otherwise would not be.
```

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+/-- A target for {name}`builtinName` and {lean}`builtinName` references. -/
+def builtinName : Nat := 42
+
+set_option doc.verso.suggestions false in
+/--
+{name}`builtinName` is a constant reference.
+
+{lean}`1 + 2 * 3` is an inline Lean term.
+
+{module -checked}`LitConfig.Core` is a module reference.
```
robsimmons:
```suggestion
{module}`LitConfig.Core` is a module reference, {module -checked}`LitConfig.Bogus` is an invalid module reference.
```

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+{tactic}`rfl` is a tactic; {conv}`rfl` is a conv tactic.
+
+{attr}`@[simp]` is an attribute.
+
+{syntaxCat}`term` is a syntax category, and {syntax term}`1 + 2` is syntax in it.
```
robsimmons:
```suggestion
{syntaxCat}`term` is a syntax category, and {syntax term}`1 + [] - (· / ·)` is syntax in it.
```

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+
+set_option doc.verso.suggestions false in
+/--
+{name}`builtinName` is a constant reference.
+
+{lean}`1 + 2 * 3` is an inline Lean term.
+
+{module -checked}`LitConfig.Core` is a module reference.
+
+{tactic}`rfl` is a tactic; {conv}`rfl` is a conv tactic.
```
robsimmons: Noting that neither of these have any hovers or apparent metadata, I think that's fine though.

test-projects/literate-config/LitConfig/UserExt.lean
```diff
@@ -0,0 +1,80 @@
...
+
+/--
+Intercepts {name}`Lean.Doc.Data.Const` docstring extensions and replaces them with a bogus
+{name (full:=VersoLiterate.Ext.data)}`data` payload. This checks that user-registered handlers are
+loaded and given precedence over built-in handlers.
+-/
+@[inline_to_literate]
+def customConst : InlineToLiterate
+  | ``Lean.Doc.Data.Const, _val, content =>
+    return some <| .other (.data "USER-CONST-MARKER") content
```
robsimmons:
```suggestion
  | ``Lean.Doc.Data.Const, _val, _content =>
    return some <| .other (.data "USER-CONST-MARKER") #[.text "(looks like you're defining a const)"]
```

makes it more obvious what's happening when looking at the non-json output

test-projects/literate-config/LitConfig/UserExt.lean
```diff
@@ -0,0 +1,80 @@
...
+
+/--
+Intercepts {name}`Lean.Doc.Data.LeanBlock` docstring extensions and replaces them with a bogus
+{name (full:=VersoLiterate.Ext.data)}`data` payload. This checks that user-registered handlers are
+loaded and given precedence over built-in handlers.
+-/
+@[block_to_literate]
+def customLeanBlock : BlockToLiterate
+  | ``Lean.Doc.Data.LeanBlock, _val, content =>
+    return some <| .other (.data "USER-LEANBLOCK-MARKER") content
```
robsimmons:
```suggestion
  | ``Lean.Doc.Data.LeanBlock, _val, _content =>
    return some <| .other (.data "USER-LEANBLOCK-MARKER") #[.blockquote #[.para #[.text "Replacement For A Lean Block"]]]
```

makes it more obvious what's happening when you're looking at the non-json output

src/tests/Tests/LiterateHtml.lean
```diff
@@ -1099,13 +1166,16 @@ private def htmlTests (data : TestData) (projectDir : System.FilePath) : List (S
   ("single-root nav flattening", testSingleRootNavFlattening data),
   ("docstrings_as_text", testDocstringsAsText data),
   ("docstrings_as_text default", testDocstringsAsTextDefault data),
-  ("keyword role", testKeywordRole data)
+  ("keyword role", testKeywordRole data),
+  ("all built-in doc roles", testAllBuiltinDocRoles data),
+  ("custom literate handlers", testCustomLiterateHandlers data),
+  ("unknown extension fallback", testUnknownExtensionFallback)
```
robsimmons, line 1205: Slightly confusing that these two tests appear to be testing the same LitConfig file, but I think it's probably fine.

test-projects/literate-config/LitConfig/Builtins.lean
```diff
@@ -0,0 +1,151 @@
...
+/-- info:
+Lean.Doc.assert
+Lean.Doc.assert'
+Lean.Doc.attr
+Lean.Doc.conv
+Lean.Doc.given
+Lean.Doc.givenInstance
+Lean.Doc.kw
+Lean.Doc.kw!
+Lean.Doc.kw?
```
robsimmons, line 117: `kw!` and `kw?` aren't exercised.

robsimmons (commented):
Sorry for the delay — took an embarrassingly long time to understand what was going on generally with the test structure, but I don't think this is the PR to make it more uniform.

Notes fall into two categories:

 - Nice-to-have improvements to tests
 - Probably-not-a-problem observations I made looking at the test files that may be separate issues/prs
 - Genuine lapses in the test file (I'm guessing Claude straight-up cheated giving "test coverage" to `assert'`

Some ambiguity between those last two categories, and the following also lands in this ambiguity gap: I naively expected this to work and it did not.

````
{open Nat}

```lean
example := succ
```
````

does this match your expectation?

If you want to schedule a quick file-tickets-or-followups session I'm happy to do that, we shouldn't delay getting this PR merged as it's a big improvement for a severe bug.

test-projects/literate-config/LitConfig/Builtins.lean
robsimmons: (the `-checked` wasn't doing anything before)

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: While this is fine to do, the test here is to check that `Add Nat` renders correctly. Tests inside of Lean already ensure that `givenInstance` registers the local instance var.

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: It's not really possible to use it after the equality type's notation has been defined. I suppose the right thing to do is to go upstream and replace it with a role that takes two or three code parameters, like ``{assert'}[`Nat.zero` `Nat.zero`]``.

test-projects/literate-config/LitConfig/Builtins.lean
robsimmons: reasonable, makes sense to decline this one

test-projects/literate-config/LitConfig/Builtins.lean
robsimmons: I recommend removing this line and just putting a comment "Untested roles: `assert'` because it's not testable currently after the equality type's notation is defined" instead.

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: Fixed!

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: Fixed!

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: This needs fixing upstream.

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: This needs fixing upstream.

david-christiansen: Fixed in https://github.com/leanprover/lean4/pull/14039

test-projects/literate-config/LitConfig/Builtins.lean
david-christiansen: Fixed in https://github.com/leanprover/lean4/pull/14039

PR #876, opened 2026-06-15
feat: lightweight HTTP server for Verso projects

david-christiansen:
Adds a lightweight HTTP server for previewing generated Verso HTML locally, run with `lake exe serve`. The server binds only 127.0.0.1, reducing the temptation to deploy it to production.

This replaces Python or Node local servers, with defaults that are better-suited to hosting Verso correctly. It also doesn't rely on users having any other programming language ecosystem available.

It can be configured from the command line, or for slightly more complex setups, via `serve.toml`.

Along the way, this branch encountered a bug in the TeX code generation that needed fixing in order to render the docs for the feature.


README.md
```diff
@@ -283,9 +283,8 @@ To use this in your project:
     lake exe verso-html .lake/build/literate html
     ```
 
-You can preview the resulting files by running
-`python3 -m http.server 8000 -d html` and pointing a web browser to
-http://localhost:8000/
+You can preview the resulting files by running `lake exe serve html`
```
robsimmons: Let's do `lake exe verso-serve` for this. (Todo for me: do the cli arguments to `verso-serve` make sense?)

robsimmons: (Check that no-args serves `.`)

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+
+```
+$ lake exe serve
+```
+
+A directory to serve and a port may be given on the command line:
+
+```
+$ lake exe serve --port 8000 _out/html
+```
```
robsimmons: Currently this is a little redundant with the previous instructions.

````suggestion
The `verso-serve` command takes an optional port (8000 by default) and an optional directory (the current directory by default).

```
$ lake exe serve --port 8000 _out/html
```
````

File: doc/UsersGuide/Serve.lean [hunk 3]
```diff
@@ -0,0 +1,293 @@
...
+open Verso.Doc
+
+#doc (Manual) "Development Server" =>
+%%%
+tag := "serve"
+htmlSplit := .never
+%%%
+
+Verso includes a small HTTP server for previewing generated HTML on your own machine.
+Run it with `lake exe serve`.
```
robsimmons: Change to `verso-serve` throughout.

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+
+# Command-Line Options
+%%%
+tag := "serve-options"
+%%%
+
+: `DIR`
+
+  The directory served at `/`, relative to the current directory. Defaults to the current directory.
+
```
robsimmons: This is confusing without a USAGE-style line that explains that DIR is a positional argument. Maybe just start "in addition to the directory served, which is given as a positional argument, the following arguments are supported:

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+  The port to listen on. Defaults to `8000`.
+
+: `--strict-port`
+
+  Fail if the port is in use instead of trying another.
+
+: `--config FILE`
+
+  Load configuration from `FILE`. Defaults to `./serve.toml` if it exists.
```
robsimmons:
```suggestion
  Load configuration from `FILE`. Defaults to `./serve.toml` if it exists. Otherwise uses defaults (see {ref "serve-config"}[Configuration file specification]).
```

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+  Disable directory listings.
+
+: `--no-trailing-slash-redirect`
+
+  Serve directories in place instead of redirecting to add a trailing slash.
+
+: `--cors`
+
+  Send permissive cross-origin headers.
+
```
robsimmons, line 64: --cors, --no-trailing-slash-redirect, and --no-listing shouldn't be command line arguments at all, they belong in serve.toml. There's a strong argument for port being configurable in both places, but I think it's best to have things configurable in one place unless there's a strong justification.

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+  Show usage and exit.
+
+Command-line options override values from the configuration file.
+
+# Configuration File
+%%%
+tag := "serve-config"
+%%%
+
+For anything beyond serving a single directory, place a `serve.toml` file next to the project, or point at one with `--config`.
```
robsimmons: "Next to the project" assumes verso-serve is being invoked in the folder with lakefile.toml, which isn't necessarily the case. 

```suggestion
For anything beyond serving a single directory, create a `serve.toml` file. When `verso-serve` is invoked, it will look for `serve.toml` in the current working directory unless another location is specified with the `--config` {ref "serve-options"}[command line option].
```

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+# Caching
+%%%
+tag := "serve-caching"
+%%%
+
+Every response carries `Cache-Control: no-cache`, which directs the browser to revalidate before reusing a cached copy rather than to skip caching entirely.
+Each file response also carries an `ETag` derived from the file's contents and a `Last-Modified` time.
+
+When a browser revalidates with `If-None-Match` or `If-Modified-Since` and the file is unchanged, the server replies `304 Not Modified` with no body.
+Because the `ETag` follows the contents, a rebuild that rewrites a file without changing it still revalidates cheaply.
```
robsimmons:
```suggestion
Because the `ETag` follows the contents, a rebuild that rewrites a file without changing it still will still be cheap to load.
```

Though this is almost 100% a consideration for a server that's serving over the network, which we're insisting this won't be...

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+For anything beyond serving a single directory, place a `serve.toml` file next to the project, or point at one with `--config`.
+Directories named in the configuration file are resolved relative to the configuration file itself.
+Every setting is optional, and an empty file behaves the same as no file.
+
+Every option is described here.
+The sections that follow add detail on mounts, redirects, and custom headers.
+
+: `port`
+
+  The port to listen on. Must be a number from `1` to `65535`. Defaults to `8000`.
```
robsimmons: If we duplicate config we should specify who wins

```suggestion
  The port to listen on. Must be a number from `1` to `65535`. Defaults to `8000`. If a port is also specified on the command line, the command-line option will be preferred.
```

robsimmons (changes requested):
Some documentation comments (which imply some code changes too, especially around removing command-line options to avoid duplication). I'm still not sure every configuration option makes sense for a development-only and/or verso-specific server, but that may be a failure of imagination — and these defaults seem reasonable.

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+
+:::paragraph
+As requests arrive, each is logged with the active port, the method, the path, and the response status:
+
+```
+[8000] GET /index.html 200
+[8000] GET /missing 404
+```
+:::
+
```
robsimmons: I think this would read better without this section — it's behavior description that will be either obvious to the user or still confusing with this explanation; it would be better to get on to the command-line options.

doc/UsersGuide/Serve.lean
```diff
@@ -0,0 +1,293 @@
...
+Every option is described here.
+The sections that follow add detail on mounts, redirects, and custom headers.
+
+: `port`
+
+  The port to listen on. Must be a number from `1` to `65535`. Defaults to `8000`.
+
+: `banner`
+
+  A line shown in the startup banner to identify the project. Must be a string. No banner is shown by default.
```
robsimmons: This feels like a pretty high documentation-noise to user-value-signal ratio. Was there a use case for a configurable command-line banner you had in mind?

src/verso-serve/VersoServe/Config.lean
```diff
@@ -0,0 +1,408 @@
...
+      else loop { acc with dir := some other } more
+  loop {} args
+
+/--
+Applies command-line overrides to a config loaded from a file.
+
+Scalar flags override the corresponding config values. A positional directory replaces the `/`
+mount. If no `/` mount remains, a default one serving the current directory is added, so `/` is
+always mapped.
+-/
```
robsimmons, line 359: The mount-juggling is strange in a function called `withCli`, and I think if there are any mounts in the config file then `lake exe verso-serve` should *not* auto-mount the root directory — and possibly should reject a positional argument altogether. The port winning on the CLI makes sense, but the mount isn't anywhere near so obvious to me. 

If the ServeConfig/CliArgs overlap is limited to mounts and ports this could probably be inlined.

README.md
robsimmons: (it does)

robsimmons (changes requested):
mostly comments on the docs / spec, i didn't review tests but the Config/Mime/Static/VersoServeMain quartet looks okay modulo questions about docs/spec.

doc/UsersGuide/Serve.lean
david-christiansen: From conversation with @robsimmons: "banner" was initially understood as HTML injection rather than command-line dump. Be more specific here.

doc/UsersGuide/Serve.lean
david-christiansen: I went with a somewhat more verbose (and grammatical) alternative.

doc/UsersGuide/Serve.lean
david-christiansen: I agree and have removed them.

doc/UsersGuide/Serve.lean
david-christiansen: `lake exe` only works from the project root, so I think this is actually the right way to do things.

doc/UsersGuide/Serve.lean
david-christiansen: Added

src/verso-serve/VersoServe/Config.lean
david-christiansen: The positional argument is now rejected when mounts are configured

doc/UsersGuide/Serve.lean
david-christiansen: Agreed - it's gone.

doc/UsersGuide/Serve.lean
david-christiansen: I think the new text addresses this

robsimmons (approved):

PR #877, opened 2026-06-16
fix: parse decoded string literals with accurate decoded positions

david-christiansen:
This fixes an issue where syntax resulting from parsed string literals would have bogus position information, which got in the way of some LSP features. Even worse, if this syntax was used in a point of interest, it panicked (after the merge of #771).


david-christiansen:
This is blocking a release, so I'll go ahead and merge.

PR #878, opened 2026-06-16
fix: use realistic source spans for Markdown

david-christiansen:
This fixes a panic in some clients of the Markdown support that used it to add blocks to a document after #771.


david-christiansen:
This is blocking a release so I'm just queuing this up.

PR #879, opened 2026-06-18
fix: parse parenthesized Markdown link URLs

### LINKED ISSUE #803: Bug: Markdown link parsing fails for URLs containing parentheses
I am working on a fork of the book Functional Programming in Lean, with Verso as the document generator. However, when it comes across URL with parentheses, i.e.

[Instance](https://en.wikipedia.org/wiki/Instance_(computer_science))

The output would be `Instance)` and the link would be `https://en.wikipedia.org/wiki/Instance_(computer_science`.

As my end user perspective, the problem would belongs to Verso. But I used LLM to read the whole repository, and it told me it was the upstream Lean 4 parser's fault. I would just raise this issue here in case my initial assumption was correct. Hopefully this could be fixed soon!

nkgotcode:
Fixes #803.

This updates Markdown inline link parsing so balanced parentheses inside URL targets remain part of the URL, while the closing delimiter and following punctuation are still parsed separately.

Validation:
- `lake env lean src/tests/Tests/Refs.lean`
- `lake test`

### DIFF
--- a/src/tests/Tests/Refs.lean
+++ b/src/tests/Tests/Refs.lean
@@ -29,6 +29,29 @@ info: Verso.Doc.Part.mk
 
 /- ----- -/
 
+#docs (.none) parenthesizedUrlLink "Parenthesized URL link" :=
+:::::::
+Here is [a link](https://en.wikipedia.org/wiki/Function_(mathematics)).
+:::::::
+/--
+info: Verso.Doc.Part.mk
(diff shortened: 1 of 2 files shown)


david-christiansen:
The current behavior is as intended. The goal of the syntax is to be simple, so rather than counting parentheses, it's just necessary to escape `)`.

I think this needs to be documented better, however!

PR #880, opened 2026-06-18
fix: label external declaration sections without h1

ejgallego:
This PR renders Manual docstring subsection labels as accessible labeled groups instead of nested `h1` elements inside external declaration bodies.

The previous HTML used `h1` tags for labels such as `Fields` and `Constructors`, which made them part of the document heading outline even though they only name declaration detail groups. This can confuse heading navigation for screen readers.

The new `Html.labeledGroup` helper renders a visible paragraph label, connects it to a `role="group"` container with `aria-labelledby`, and lets downstream renderers reuse the same accessibility pattern without hand-rolling it. Blueprint has already validated this shape downstream with Firefox accessibility tooling.

Thanks to @NicolasRouquette for the original bug report: leanprover/verso-blueprint#130

### DIFF
--- a/src/tests/Tests/VersoManual/Html.lean
+++ b/src/tests/Tests/VersoManual/Html.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Author: David Thrane Christiansen
 -/
 import VersoManual.Html
+import VersoManual
 
 namespace Verso.Genre.Manual.Html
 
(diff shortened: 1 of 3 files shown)


src/tests/Tests/VersoManual/Html.lean
```diff
@@ -112,3 +113,65 @@ Done
...
+
+{docstring SectionFixture}
+
+:::::::
+
+private def hasSubstring (s sub : String) : Bool :=
+  s.find? sub |>.isSome
+
+private def compactHtml (s : String) : String :=
+  String.ofList <| s.toList.filter (fun c => !c.isWhitespace)
```
david-christiansen: This is a pretty wild way to implement this function!

Why not do something like [this](https://live.lean-lang.org/#codez=CYUwZgBAdg9gzgBwIYGMQQBRwgLggZQBcAnASygHMBKXAk8i3AXgjgDowYAbYLzc0oWYQARCJpgArlABQECAB8IMSYQA0EAORaITAHzLVcxYfUQUugysJsEkuAAtzMmQHoAtO4jlOeEQB4kB2JwJgAdEU4YCL0ADwBPAC8kACMUf1ckPREId1cZAGIKSSRiYAB9AFs4CmxyQpAANyQ%2BWERUdACkCGDQiKiYhMSIVPTM7KA) instead of allocating a linked list?

```
def nospace (s : String) : String := s.foldl (init := "") fun
  | out, ' ' => out
  | out, c => out.push c

/-- info: "<ahref=\"foo\">xyzabc</a>" -/
#guard_msgs in
#eval nospace "<a href=\"foo\">xyz abc</a>"
```

src/tests/Tests/VersoManual/Html.lean
```diff
@@ -112,3 +113,65 @@ Done
...
+  let definitionIds := state.definitionIds ctxt
+  let remotes : Multi.AllRemotes := {}
+  let linkTargets := cfg.linkTargets state remotes
+  let (html, _) ←
+    (Manual.toHtml (m := ReaderT Multi.AllRemotes (ReaderT ExtensionImpls (BuildLogT IO)))
+        {} ctxt state definitionIds linkTargets {} part).run {}
+      |>.run remotes
+      |>.run extension_impls%
+      |>.run logger
+  unless (← logger.exitCode) == 0 do
```
david-christiansen: We have a helper for this pattern: https://github.com/leanprover/verso/blob/113c54a466d14d1cd1e615103aefd27ffef19b12/src/verso/Verso/BuildLog.lean#L205-L208

src/tests/Tests/VersoManual/Html.lean
```diff
@@ -112,3 +113,65 @@ Done
...
+    throw <| IO.userError "Fields section should render as a named group"
+  let label :=
+    "<pclass=\"docstring-section-label\"id=\"docstring-section-Fields\">Fields</p>"
+  unless hasSubstring compact label do
+    throw <| IO.userError "Fields section should use a paragraph label with the group's ID"
+  if hasSubstring compact "<h1>Fields</h1>" then
+    throw <| IO.userError "Fields section label should not render as h1"
+  IO.println "docstring section labels render as labeled groups"
+
+end DocstringSectionRegression
```
david-christiansen:
```suggestion

```

david-christiansen:
This looks basically right. Why is it still a draft?

ejgallego:
> This looks basically right. Why is it still a draft?

Because I still was not satisfied with the code (which in some cases was generated by Codex) and wanted to refine / review more.

src/tests/Tests/VersoManual/Html.lean
ejgallego: Thanks, fixed!

src/tests/Tests/VersoManual/Html.lean
ejgallego: Thanks, fixed!

ejgallego:
Ready for review and hopefully merge.

PR #881, opened 2026-06-18
refactor: configurable source for highlighting

robsimmons:
This PR allows doc highlighting information to be be provided from a different source, instead of just a hardcoded `-verso-docs.json` resource file. It also incidentally turns the get-highlighting-json operation from a chained promise to an await-ed one.

The reason for this change is to eventually allow the lean4web app to generate a single-file page that only refers to external static assets — the main per-document assets for a single-page HTML development are index.html, the highlighting information file, and search data (which can be omitted from a single-page HTML development).

(Hint: the diff is quite small if you preview with whitespace changes off.)

### DIFF
--- a/src/verso/Verso/Code/Highlighted.lean
+++ b/src/verso/Verso/Code/Highlighted.lean
@@ -1355,7 +1355,7 @@ Some CSS frameworks customize details/summary in ways not compatible with Verso'
 "
 
 /--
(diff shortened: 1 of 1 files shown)


david-christiansen:
This seems fine - thanks for the whitespace hint!

It would be nice if the PR description included the motivation for the change, because that will be the commit message in the future, and this could be mysterious without a bit of context. Otherwise, LGTM

PR #891, opened 2026-06-26
fix: report expected content when allowDiff > 0

robsimmons:
Currently, 

`````lean4
#docs (Genre.Manual) allowDiff31 "Not enough allowDiff" :=
:::::::
```lean (name := foo)
#print "a\nb\nc"
```
```leanOutput foo (allowDiff := 1)
b
```
:::::::
`````

will report 

```
error: Didn't match - got:
  []
but expected:
  b
```

and I can imagine a multitude of fancy ways to do that right, but that this PR does is the lazy way of piggybacking off of the existing reporting. 

```
error: Didn't match within even with allowDiff := 1 - got:
  [a
   b
   c]
but expected:
  b
```

The new error message would have gotten me from getting into a cul-de-sac trying to merge refman main into refman nightly earlier today, and now there's a test for it. (Three tests, two already passed, but the middle one failed before and is fixed by this PR.)

### DIFF
--- a/src/tests/Tests/LeanCode.lean
+++ b/src/tests/Tests/LeanCode.lean
@@ -131,3 +131,66 @@ The binding can be removed (if unused) or named `_` (if used implicitly).
 Note: This linter can be disabled with `set_option linter.unusedVariables false`
 ```
 :::::::
+
+/--
+error: Didn't match - got: ⏎
+  [a
+   b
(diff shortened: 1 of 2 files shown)


david-christiansen (approved):
This is a nice solution, thanks! That was a truly terrible error message.

PR #893, opened 2026-06-26
chore: adapt to removal of 'Name' from dynamic extension types

robsimmons:


src/verso-literate/VersoLiterate/Basic.lean
```diff
@@ -169,46 +150,37 @@ def handleConvTactic : InlineToLiterate
...
 
 def handleKwAtom : InlineToLiterate
-  | name, _val, content => do
+  | val, content => do
     -- Data.Atom is mistakenly marked private in Lean. Here's a workaround until we fix that.
     -- Check the name's suffix because private names have a mangled prefix:
-    unless name.toString.endsWith "Lean.Doc.Data.Atom" do return none
+    unless val.typeName.toString.endsWith "Lean.Doc.Data.Atom" do return none
     let some s := (match content with | #[.code s] => some s | _ => none) | return none
     return some <| .other (.highlighted <| .token ⟨.keyword none none none, s⟩) content
```
robsimmons, line 154: These two cases were the only ones where I wasn't totally confident what I was doing: the first because I'm not sure I fully understand the comment and I don't know if both cases are actually possible, and the second because the `toString` conversion and postfix check seemed pretty weird to me.

robsimmons:
!consistent

david-christiansen (approved):

PR #896, opened 2026-06-27
refactor: remove impossible cases in handler logic

robsimmons:
it would be slightly better to handle this on `main`, but due to the adaptations I applied in #893 they'd all edit-conflict with nightly-testing.


david-christiansen:
I'm not sure why this is needed - is it to make the future merge easier? Or just so that the behavior from nightly is already anticipated now?

In any case, it seems to be easy to read and match the in-practice behavior.

PR #899, opened 2026-06-28
feat: render KaTeX math in literateHtml output

robsimmons:
Adds KaTeX math to `:literateHtml` output.

For this example:

```
set_option doc.verso true
/-!
Hello world: $`x + 4`

$$`\int_0^k f(x)\,dx`
-/
````

The output of `lake build :literateHtml` looks like this on `main`

image

And with the changes in this PR it looks like this:

image

### DIFF
--- a/src/verso-literate-html/LiterateHtmlMain.lean
+++ b/src/verso-literate-html/LiterateHtmlMain.lean
@@ -167,7 +167,7 @@ open Verso Output Html Search in
 Builds the `<head>` contents for a literate page. When `includeCodeAssets` is true,
 includes popper/tippy/highlighting/copy-button assets needed for code hover tooltips.
 -/
-private def mkHeadContents (litConfig : LiterateConfig) (includeCodeAssets : Bool := true) (includeKatexAssets : Bool := true) : Html :=
+private def mkHeadContents (litConfig : LiterateConfig) (includeCodeAssets : Bool := true) (includeKaTeXAssets : Bool := true) : Html :=
   let faviconTag : Html := match litConfig.metadata.favicon with
     | some fav => {{<link rel="icon" href={{(System.FilePath.fileName fav).getD fav}}/>}}
     | none => {{<link rel="icon" href="data:,"/>}}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/verso-literate-html/LiterateHtmlMain.lean [hunk 1]
```diff
@@ -481,6 +488,15 @@ def main (args : List String) : IO UInt32 := do
...
     IO.FS.writeFile (config.outputDir / "literate.css") literate.css
+    -- Copy KaTeX dependencies
+    IO.FS.createDirAll (config.outputDir / "katex")
+    IO.FS.writeFile (config.outputDir / "katex" / "katex.css") katex.css
+    IO.FS.writeFile (config.outputDir / "katex" / "katex.js") katex.js
+    IO.FS.writeFile (config.outputDir / "katex" / "math.js") math.js
+    for (name, contents) in katexFonts do
+      let path := config.outputDir / name
+      path.parent.forM IO.FS.createDirAll
+      IO.FS.writeBinFile path contents
```
david-christiansen, line 499: Should this also be gated by `includeKaTeXAssets?`

src/verso-literate-html/LiterateHtmlMain.lean
```diff
@@ -166,7 +167,7 @@ open Verso Output Html Search in
 Builds the `<head>` contents for a literate page. When `includeCodeAssets` is true,
 includes popper/tippy/highlighting/copy-button assets needed for code hover tooltips.
 -/
-private def mkHeadContents (litConfig : LiterateConfig) (includeCodeAssets : Bool := true) : Html :=
+private def mkHeadContents (litConfig : LiterateConfig) (includeCodeAssets : Bool := true) (includeKatexAssets : Bool := true) : Html :=
```
david-christiansen:
```suggestion
private def mkHeadContents (litConfig : LiterateConfig) (includeCodeAssets : Bool := true) (includeKaTeXAssets : Bool := true) : Html :=
```

I think that when we can reasonably do so, it's good to follow the project's capitalization.

src/verso-literate-html/LiterateHtmlMain.lean
robsimmons: This output is per-project — we just avoid including katex (and code) assets on the landing page.

Would it make sense to you for me to add a note "if you make rendering katex conditional per-module depending on whether math is used, then you should avoid outputting katex at all if it's never used in any module," so we don't miss this in the future?

src/verso-literate-html/LiterateHtmlMain.lean
robsimmons: done

src/verso-literate-html/LiterateHtmlMain.lean
david-christiansen: Right - I read too quickly!

I suppose the note might be useful but I don't think it's a big deal either way.

david-christiansen (approved):

PR #900, opened 2026-06-29
fix: add braces to item description to avoid errors on `]`

robsimmons:
In writing #876, @david-christiansen noted "this branch encountered a bug in the TeX code generation that needed fixing" — but I think the issue was slightly misdiagnosed, it wasn't a fragility issue, a `]` anywhere in an item description (including outside of verbatim) caused a rendering issue because it was read as the description closer.

This fix fixes the bug from #876, also fixes bugs for `]` in literal text, and adds a test for the behavior.


david-christiansen:
Ah, so going from verbatim mode to `texttt` solved the issue as a side effect by introducing escaping?

robsimmons:
yes — but we don't escape regular text, so this verso

```
* This parenthetical [will

   lorum ipsum 

* cross labels]

   lorum ipsum  
```

misrendered in TeX with your fix

PR #902, opened 2026-07-01
feat: proper test framework

david-christiansen:
This PR rearchitects our test suite into a proper testing framework.

Before, our tests were all essentially _ad hoc_. Each kind of test would be an IO action that might throw, and there were some rough conventions, but output reporting was inconsistent and sometimes hard to understand. Tests were noisy as well, which made it hard to find reasons for failure at a glance.

Now, there's a test framework called `Errata`, designed to be something we can extract to its own repo after we get some experience with it here (so it can be used e.g. in `verso-slides` and `lean-sqlite`).

In `Errata`, tests are marked by the `@[test]` attribute. Their docstring and source range are saved for failure reporting. The value of a test can have any type with an `IsTest` instance. Some machinery in the Lakefile enumerates all tests, providing them to the runner.

Elaboration-time tests can also be implemented using `#test_msgs` and `#test_guard`, which are versions of `#guard_msgs` and `#guard` that run the compile-time test but save the result as a test case for reporting together with the rest of the tests.

Errata also supports saving JUnit XML, which various GitHub actions can conveniently display for us.

Tests can additionally be run interactively using a widget.


david-christiansen:
Some examples of the widget:

image

It also captures output and provides timestamps and indicates whether it was stdout or stderr:
image

david-christiansen:
@tydeu If you could check out the Lake parts of this and make sure I'm not doing anything that's likely to fail unexpectedly, it'd be appreciated. Thanks!

lakefile.lean
```diff
@@ -124,15 +124,256 @@ lean_exe «verso-literate-plan» where
...
+Whether a source file introduces Errata tests, by an `@[test]` attribute (applied inline or with a
+separate `attribute [test] …`), a `#test_msgs` command, or a `#test_guard` command. This drives the
+glob-coverage check, which reads source files before anything is built; test discovery itself reads
+the compiled modules.
+-/
+private def errataSourceHasTests (lines : List String) : Bool :=
+  lines.any fun line =>
+    let t := line.trimAsciiStart.copy
+    t.startsWith "@[test]" || t.startsWith "attribute [test" ||
+      t.startsWith "#test_msgs" || t.startsWith "#test_guard"
```
tydeu, line 204: This seems very fragile. Perhaps one could (also?) mark errata files with a line like `--! errata`?

lakefile.lean
```diff
@@ -124,15 +124,256 @@ lean_exe «verso-literate-plan» where
...
+      [("selection", selection ++ "\n"),
+       ("ErrataDiscovered.lean", errataDiscoveredSource ws.root.prettyName moduleMods),
+       ("ErrataRunnerMain.lean", errataMainSource ws.root.prettyName nonModuleMods `ErrataDiscovered)] do
+    let file := dir / name
+    let changed ← if ← file.pathExists then pure ((← IO.FS.readFile file) != src) else pure true
+    if changed then IO.FS.writeFile file src
+  -- Build and run the discovered runner.
+  let some exe := ws.findLeanExe? `«errata-runner»
+    | IO.eprintln "errata-runner executable is not configured"; return 1
+  let exePath ← runBuild exe.fetch
```
tydeu:
```suggestion
  let exePath ← runBuild «errata-runner».fetch
```

This should work (and be more forward-compatible).

lakefile.lean
```diff
@@ -124,15 +124,256 @@ lean_exe «verso-literate-plan» where
...
+  let modInfos ← runBuild do
+    let mut oleanJobs := #[]
+    let mut infos : Array (Lean.Name × System.FilePath) := #[]
+    for lib in libs do
+      let mods ← (← lib.modules.fetch).await
+      for m in mods do
+        oleanJobs := oleanJobs.push (← m.olean.fetch)
+        infos := infos.push (m.name, m.oleanFile)
+    let _ ← (Job.collectArray oleanJobs).await
+    pure (Job.pure infos)
```
tydeu:
```suggestion
    (Job.collectArray oleanJobs).map (sync := true) fun _ => infos
```

Avoids blocking a thread for the jobs to finish.

lakefile.lean
```diff
@@ -124,15 +124,256 @@ lean_exe «verso-literate-plan» where
...
+  srcDir := ".lake/errata-runner"
+  roots := #[`ErrataDiscovered]
+  needs := #[errataSelection]
+
+-- The generated, discovered test runner. Its source is written by the Errata test driver.
+lean_exe «errata-runner» where
+  root := `ErrataRunnerMain
+  srcDir := ".lake/errata-runner"
   supportInterpreter := true
+  needs := #[errataSelection]
```
tydeu, line 192: The current workflow is to for these to fail to build until `errata-test` is run to generate them. I presume the reason for this is that they can change based on the arguments passed to `errata-test`? Otherwise, I would have suggested to use `needs` to depend on custom target which generates them. That way, e.g., `lake exe errata-runner` works even if a `errata-test` has yet to be called. It may be worth documenting the rationale for the current design in a brief comment.

tydeu (commented):
The Lake code looks reasonable to me! The comments I left are just minor suggestions. Very cool tool!

.github/workflows/test-imports.yml
```diff
@@ -1,3 +1,8 @@
+# This check is retained only to satisfy the required "Check all test modules are imported" status
+# check on the protected branch. It is a no-op on this layout: test modules now live under
+# src/tests/VersoTests and are discovered by globbing, and src/tests/Tests no longer exists, so the
+# scan below finds nothing. Coverage is instead enforced by `errataWarnUncoveredTestModules` during
+# `lake test`. Remove this workflow (and drop the required check) once merged.
```
robsimmons, line 5: Why don't we just delete the required status and then go ahead and delete this file in the PR? Seems preferable and more atomic.

src/errata/Errata/usage.txt
```diff
@@ -0,0 +1,21 @@
...
+  lake test                                         run every test in the package
+  lake test -- LIBRARY...                           run the tests in the given libraries
+  lake test -- LIBRARY... --test-options OPTION...  pass runner options after the marker
+
+Runner options go after `--test-options`; tokens before it name libraries. A library is a bare
+`Library` in this package or a `package/Library` reaching into a dependency.
+Runner options:
+  -v, --verbose        Also report passes (truncating each test's results); repeat (-vv) for all,
+                       -vvv to also show every test's docstring, not only those that fail.
+      --update-golden  Rewrite golden expected files instead of comparing.
```
robsimmons, line 13: It's weird to see this called "golden", that may just be something I'm unfamilari with though — isn't "snapshot" more standard?

src/errata/Errata/usage.txt
```diff
@@ -0,0 +1,21 @@
+Errata test runner
+
+Usage:
+  lake test                                         run every test in the package
+  lake test -- LIBRARY...                           run the tests in the given libraries
+  lake test -- LIBRARY... --test-options OPTION...  pass runner options after the marker
```
robsimmons, line 6: Strong request: can we use `-lLIBRARY`/`--lib=LIBRARY` so we're not sad when we need to add a second kind of CLI configuration?

Weak request: can we use the more standard `--` divider for runner options? `--test-options` doesn't present itself as being any kind of meta-cli-arg.

src/errata/Errata/CompileTime/Helpers.lean
```diff
@@ -0,0 +1,48 @@
...
+/-- Whether the expected and actual message blocks match, normalizing whitespace as `#guard_msgs` does. -/
+def messagesMatch (expected actual : String) : Bool :=
+  let norm := fun s => (WhitespaceMode.normalized.apply s).trimAscii.copy
+  norm expected == norm actual
+
+/-- The doc comment that would make the expected block match the actual output. -/
+def suggestedDoc (actual : String) : String :=
+  if actual.isEmpty then ""
+  else if actual.contains '\n' then s!"/--\n{actual}\n-/\n"
+  else s!"/-- {actual} -/\n"
```
robsimmons, line 48: All of these are waiting-to-shoot-us-in-the-foot functions that really ought to be made public and factored out of `Lean.Elab.GuardMsgs`

src/errata/Errata/widget/jsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+        "lib": ["ES2024", "DOM", "DOM.Iterable"],
+        "target": "ES2024",
+        "noEmit": true
```
robsimmons, line 5:
```suggestion
        "noEmit": true,
        "strict": false
```

Strict is true by default now ([typescript 6](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#simple-default-changes)) and in recently-updated VSCode versions this will show an error (because VSCode generally uses most-recent typescript) unless we specify strict:false

src/errata/Errata/Assertions.lean
```diff
@@ -0,0 +1,69 @@
...
+
+set_option linter.missingDocs true
+set_option doc.verso true
+
+namespace Errata
+
+/-- Asserts that a condition holds. -/
+def assert (cond : Bool) (message : String := "assertion failed")
+    (loc : Location := by exact here%) : TestM Unit :=
+  unless cond do failAt loc message
```
robsimmons:
```suggestion
def assert (cond : Bool)
    (loc : Location := by exact here%) : TestM Unit :=
  unless cond do failAt loc "assertion failed"
```

Series of yagni complexity nitpicks here — made worse by the fact that *some* asserts don't take messages, and it should definitely be consistent one way or the other

src/errata/Errata/Assertions.lean
```diff
@@ -0,0 +1,69 @@
...
+def assertContains (expected actual : String) (message : String := "substring not found")
+    (loc : Location := by exact here%) : TestM Unit := do
+  unless (actual.find? expected).isSome do
+    failAt loc message (detail? := some s!"expected to contain: {expected}\nactual: {actual}")
+
+/-- Asserts that the actual string does not contain the unexpected substring. -/
+def assertNotContains (unexpected actual : String) (message : String := "unexpected substring found")
+    (loc : Location := by exact here%) : TestM Unit :=
+  unless (actual.find? unexpected).isNone do
+    failAt loc message (detail? := some s!"expected not to contain: {unexpected}\nactual: {actual}")
```
robsimmons:
```suggestion
def assertNotContains (unexpected actual : String)
    (loc : Location := by exact here%) : TestM Unit :=
  unless (actual.find? unexpected).isNone do
    failAt loc "unexpected substring found" (detail? := some s!"expected not to contain: {unexpected}\nactual: {actual}")
```

src/errata/Errata/CompileTime.lean
```diff
@@ -0,0 +1,149 @@
...
+
+set_option doc.verso true
+
+/--
+When true, a failing Errata compile-time test is an elaboration error rather than a warning.
+-/
+register_option errata.failOnError : Bool := {
+  defValue := false
+  descr := "Make a failing Errata compile-time test an elaboration error rather than a warning."
+}
```
robsimmons, line 25: Not a demand by any means, but I do kind of subscribe to the "don't have warnings" school of thought. Why not just start with failing Errata compile-time tests as elaboration errors? Or is the idea here that the tests are "really" being run separately and the elaboration warnings are just hints... in which case do we really need to start with that configurability knob?

src/errata/Errata/widget/run_test_widget.js
```diff
@@ -0,0 +1,515 @@
...
+        chunks.length === 0
+            ? null
+            : e(
+                  "details",
+                  {
+                      key: "output",
+                      open: outputOpen,
+                      onToggle: function (ev) {
+                          setOutputOpen(ev.target.open);
+                      },
```
robsimmons:
```suggestion
                      onToggle: /** @param ev {React.ToggleEvent<HTMLDetailsElement>} */ function (
                          ev,
                      ) {
                          setOutputOpen(ev.currentTarget.open);
                      },
```

fixes a TS error that would occur if we were more honestly using React types.

src/errata/Errata/widget/widget-externals.d.ts
```diff
@@ -0,0 +1,16 @@
...
+    export function useState(initial: any): [any, (value: any) => void];
+    export function useEffect(effect: () => void | (() => void), deps?: any[]): void;
+    export function useRef(initial: any): { current: any };
+}
+
+declare module "@leanprover/infoview" {
+    export function useRpcSession(): {
+        call(method: string, params: any): Promise<any>;
+    };
+}
```
robsimmons, line 20: I'd suggest replacing this whole file with

```
declare module "@leanprover/infoview";
```

We're not using TS in a way that gives us effectively any guarantees here, so having an easily-capable-of-drift .d.ts file here is mostly security theater. 

Removing the `react` module would require updating `package.json`; I can do this in another PR if you prefer but I think it should be the following:

```json
{
    "devDependencies": {
        "@types/react": "^19.2.17",
        "prettier": "3.9.4",
        "typescript": "6.0.3"
    }
}
```

robsimmons (changes requested):
Have some minor comments, but ran into two larger comments:

 - The Assertions module doesn't consistently allow a user defined message, that API feels like it should be uniform
 - I think we want the Assertions module to consistently define a message that is (Option (Unit -> String)) rather than a string — the user might want to give a computationally-expensive error message and that's waiting to cause us pain with the current API.

I could nitpick the widget to death and would somewhat prefer to ship an MVP without it, but realistically we could switch it out later as long as the RPC protocol is sound.

src/errata/Errata/widget/widget-externals.d.ts
Vtec234: If bumping `package.json`, you can also add `@leanprover/infoview` as a dev dependency and remove `widget-externals.d.ts` altogether.

src/errata/Errata/Widget.lean
```diff
@@ -0,0 +1,283 @@
...
+set_option linter.missingDocs true
+set_option doc.verso true
+
+open Lean
+
+namespace Errata.Widget
+
+/--
+The InfoView widget shown when the text cursor is on a test's {lit}`@[test]` marker. It offers a Run
+button that runs the test in the language server, streaming its output as it is produced.
```
Vtec234:
```suggestion
Shown when the text cursor is on a test's source span.
Offers a Run button that runs the test in the language server,
streaming its output as it is produced.
```
Not just the marker!

src/errata/Errata/Widget.lean
```diff
@@ -0,0 +1,283 @@
...
+/--
+Builds the test's module from the saved source, then runs the test, streaming its output into the
+run state. Building first means a Run reflects the latest saved version of the test.
+-/
+private meta def buildAndRun (module declJson : String) (state : RunState) : IO Unit := do
+  -- `lake query` builds the runner exe and the test's module (so the run reflects the saved source)
+  -- and prints the exe's absolute path on stdout; progress and errors go to stderr.
+  let build ← IO.Process.spawn {
+    stdin := .null, stdout := .piped, stderr := .piped
+    cmd := "lake", args := #["query", "errata-run-one", module]
```
Vtec234, line 176: I don't understand the setup but this looks sus; we are already in the language server, why are we running Lake instead of getting the test IR from the environment and interpreting it?

lakefile.lean
david-christiansen: Hmm, maybe it'd be better just to delete the warning?

The situation I want to avoid is someone saying "I thought my tests were running!" when they aren't. I think remembering to add `--! errata` is no more likely than remembering to change the glob value, so I don't think it would catch this kind of thing.

But this is indeed fragile. Probably deleting it is better, and instead warning if the globs setting of a library that contains any test at all is the wrong one, or just documenting the footgun loudly in a README. I'll sleep on it.

src/errata/Errata/usage.txt
david-christiansen: "Golden testing" is a fairly established term for "run the program and check that the file it emits is identical to this known-good one".

.github/workflows/test-imports.yml
david-christiansen: OK. When we decide "merge", I can delete it from the PR and remove it from the required set. Until then, let's keep things passing.

src/errata/Errata/usage.txt
david-christiansen: The unit of execution here for the runner is a library. This seems a bit like wanting `cp --source=a.txt --dest=b.txt` instead of `cp a.txt b.txt` in case `cp` wants something more.  

It had the `--` for the second divider before. I found myself repeatedly making the mistake that I'd write something like `lake test -- --enable-tex` instead of `lake text -- -- --enable-tex`, and I stopped making that mistake when the second `--` became  `--test-options`. This is not a strongly-held preference, but it did arise from working with the thing.

src/errata/Errata/widget/jsconfig.json
david-christiansen: Should it be false? Or should it be true, and the code improved to match?

src/errata/Errata/widget/run_test_widget.js
```diff
@@ -0,0 +1,515 @@
...
+}
+
+export default function (props) {
+    const rs = useRpcSession();
+    // Keyed by both the test and a hash of its source, so editing the test changes the key and
+    // invalidates its cached/in-progress run.
+    const version = props.version || "";
+    const cacheKey = JSON.stringify(props.decl) + "@" + version;
+
+    const [outcome, setOutcome] = React.useState(() => resultCache.get(cacheKey) || null);
```
Vtec234, line 118: I usually prefer using state machines that disallow nonsense states; a long sequence of `useState`s means that we can simultaneously have a successful `outcome` and `error !== null`, etc. Downstream from that, some of the effects mutate part of the run state incorrectly, e.g. `startedAt.current` is reset to `Date.now` if you click off and back into a test while it's still running. But maybe we're happy with something that works well enough 🤷

src/errata/Errata/CompileTime.lean
david-christiansen: The thing that's funny about Lean testing is that we usually have tests at two separate phases: elaboration-time and run-time. Most languages have only the second.

The goal with this setup is to give a single list of all tests and their results and make the interface as uniform as it can be. If elaboration-time tests block elaboration, then a failing test prevents all downstream test modules from running at all, and there's no way for it to be included in an overall report of passes/fails. So the test library's version of `#guard_msgs` and friends save their test result in the environment as essentially a constant test, already defined as a pass or a fail, so it can be included in the run-time-phase test execution. This gives uniformity.

It also has the annoying consequence that, while editing the test, useful information is hidden - the fact that it fails! Logging a warning surfaces this so that the author can see to fix it. The idea is that these warnings never get actually committed to `main`, but they can exist while working on a feature and tracking progress.

I'd be up for deleting this option.

src/errata/Errata/widget/widget-externals.d.ts
david-christiansen: Would you be up for just making this whole TypeScript situation be actually good for this widget in such a way that users don't need `npm` if they don't want to replicate what CI is doing? I'm really a n00b here, and I think you could make it excellent in like 5 minutes. If you're up for it, please just push to the branch!

src/errata/Errata/Widget.lean
david-christiansen: Argh, that's how it used to be, and then I improved it and forgot to fix the docs. Thanks!

src/errata/Errata/Widget.lean
david-christiansen: It's because we don't really have a way to kill a task in Lean. Tasks need to check for cancellation requests, and arbitrary tests don't typically do that. The fear is someone clicking "run" then "stop" then "run" then "stop" then ..., and ending up starving the language server's thread pool.

Running a process, on the other hand, gives us a way to kill it.

lakefile.lean
david-christiansen: You understand correctly. I'll do the comment.

david-christiansen:
> * The Assertions module doesn't consistently allow a user defined message, that API feels like it should be uniform

I agree and will fix.
 
> * I think we want the Assertions module to consistently define a message that is (Option (Unit -> String)) rather than a string — the user might want to give a computationally-expensive error message and that's waiting to cause us pain with the current API.

That's true, but also annoying!

What about something like:

```
structure AssertionMessage where
  text : Unit -> String
deriving blah blah blah

instance : Coe String AssertionMessage where
  coe str := { text := fun () => str }
```

I think that lets you write the string literal that is wanted in almost all cases, and also delays it. Or maybe have a sum type with an immediate case and a closure case with strings coercing to the immediate one, if we're worried about too many closures.

src/errata/Errata/widget/run_test_widget.js
david-christiansen: I'd like it to be actually good!

src/errata/Errata/widget/widget-externals.d.ts
david-christiansen: (here "you" was meant to be @robsimmons but @Vtec234 is also very welcome to do it)

robsimmons: I'm afraid that without `npm` there's just no way for typescript to get the types for react or infoview or etc. I can experiment with whether vscode currently is able to figure them out without an npm install, but I'm not optimistic.

src/errata/Errata/CompileTime.lean
robsimmons: I agree with your rationale for the warning — but I think failOnError is worth deleting unless/until there's a compelling reason for it, and it sounds like maybe you agree.

src/errata/Errata/widget/widget-externals.d.ts
david-christiansen: I think it's totally fine to run `npm` in CI, and as an option for local use when actually working on the code so you don't have to wait for CI. I just don't want our _users_ to need it in order to use the widget or run tests - hence "TypeScript in comments" instead of proper TypeScript.

src/errata/Errata/NameJson.lean
```diff
@@ -0,0 +1,38 @@
...
+set_option linter.missingDocs true
+set_option doc.verso true
+
+namespace Errata
+
+open Lean
+
+/--
+Encodes a {name}`Lean.Name` structurally, preserving the numeric and hygienic components that the
+standard string form does not round-trip. The widget and the single-test runner exchange test names
```
Vtec234, line 21: What are the non-round-tripping examples?

src/errata/Errata/Runner.lean
```diff
@@ -0,0 +1,184 @@
...
+    else if arg.startsWith "-" && arg.length > 1 then
+      -- Expand bundled short flags: `-xyz` becomes `--x --y --z`.
+      let expanded := (arg.drop 1).copy.toList.map (fun c => "--" ++ toString c)
+      rawOptions (expanded ++ rest)
+    else
+      .error s!"unexpected argument: {arg}"
+
+/-- Parses the runner's command-line arguments, peeling off known flags and keeping the rest as
+project options. -/
+def parseArgs (args : List String) : Except String Options := do
```
Vtec234, line 129: Could we just use https://github.com/leanprover/lean4-cli?

david-christiansen:
Superseded by #956, #957, #959

jcreedcmu:
Closing in favor of the stack:
https://github.com/leanprover/verso/pull/956
https://github.com/leanprover/verso/pull/957
https://github.com/leanprover/verso/pull/959

PR #906, opened 2026-07-06
chore: drop no-op `@[expose]` on `optionDocs.Args`

kim-em:
The `@[expose]` attribute on `optionDocs.Args` in `src/verso-manual/VersoManual/Docstring.lean` has no effect because that file is not a `module` file, and it produces a `@[expose] has no effect outside a module file` warning on every build (including in downstream projects that build the manual). This removes the attribute; it can be reintroduced if and when the file is migrated to the module system.

The change is a no-op for compilation (the compiler itself reports the attribute has no effect), so behaviour is unchanged.


robsimmons:
This may have been a skew with nightly-testing; I'll make sure it survives in nightly-testing if that's the case

PR #908, opened 2026-07-07
chore: factor tests into a different file

robsimmons:
The partial module-ification of nightly is making the combined Html file difficult for me to work with, this conceptually simplifies what I'm doing a little bit.


robsimmons:
fyi @ejgallego just in case you wonder where these tests went, it made my life slightly easier to move them

PR #913, opened 2026-07-09
fix: set left and right margins to be equal in TeX output

medovina:
A TeX file generated by Verso uses the `memoir` documentclass.  The default layout of this documentclass has left and right margins that differ greatly.  That might be OK for printing, but it is jarring when you view a PDF file in a desktop application, which is the typical use case today.

This commit modifies the left and right margins to be equal.  In my opinion this makes the PDF quite a bit nicer to view in Evince (on Linux), and probably in other PDF viewers too.

### DIFF
--- a/src/verso-manual/VersoManual/TeX.lean
+++ b/src/verso-manual/VersoManual/TeX.lean
@@ -9,6 +9,8 @@ namespace Verso.Genre.Manual.TeX
 public def preamble (title : String) (authors : List String) (date : String) (packages : List String) (extraPreamble : List String) : String :=
 r##"
 \documentclass{memoir}
+\setlrmarginsandblock{1.7in}{1.7in}{*}
+\checkandfixthelayout
 
 \usepackage{sourcecodepro}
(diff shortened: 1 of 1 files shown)


src/verso-manual/VersoManual/TeX.lean
```diff
@@ -9,6 +9,8 @@ namespace Verso.Genre.Manual.TeX
 public def preamble (title : String) (authors : List String) (date : String) (packages : List String) (extraPreamble : List String) : String :=
 r##"
 \documentclass{memoir}
```
david-christiansen, line 11: Shouldn't this be accomplished by passing the `oneside` option to `memoir`?

In any case, what about making this into a configurable option that defaults to single-sided printing but can pass `twoside` if requested for those who actually do print double-sided and want the recto/verso margin distinction for the binding?

medovina: Yes, that would be a better approach.  I'll close this pull request and work on making a new one with a configurable option as you suggested.

PR #914, opened 2026-07-09
fix: try removing with-deps from Playwright CI

robsimmons:


david-christiansen:
What does this fix?

I don't remember why that's there, so it's hard to evaluate whether this is needed.

robsimmons:
This was an attempt to fix something in downstream-lean4, we can ignore this for now

Issue #918, opened 2026-07-09
paragraphs are not preserved in docstring boxes in TeX output


medovina:
In docstring boxes in TeX output, all paragraphs are combined into a single paragraph, making the text difficult to read.

For example, section 3.3 Traversal of the Verso documentation contains a docstring box describing the Verso.Doc.Traverse type class.  In the generated HTML, its description looks like this:

> Genre-specific traversal.
>
> The traversal pass is where cross-references are resolved. The traversal pass repeatedly applies a genre-specific stateful computation until a fixed point is reached, both with respect to the state and the document. Traversal may update the state or rewrite parts of the document.
>
> The methods part, block, and inline provide effects to be carried out...

But in the TeX output all of these paragraphs are combined, so the resulting PDF looks like this:

> Genre-specific traversal.The traversal pass is where cross-references are re-
solved. The traversal pass repeatedly applies a genre-specific stateful compu-
tation until a fixed point is reached, both with respect to the state and the doc-
ument. Traversal may update the state or rewrite parts of the document.The
methods part, block, and inline provide effects to be carried out...

PR #919, opened 2026-07-09
fix: preserve paragraphs in docstring boxes in TeX output (#918)

medovina:


david-christiansen:
Can you include a quick screenshot here? That'd make this much faster to review!

medovina:
Here are two PDF files of the Verso documentation itself, one built from master (main_master.pdf), the other built with this fix (main_fix.pdf).  I suggest looking e.g. at section 3.3 Traversal in both documents, especially at the box describing the Verso.Doc.Traverse type class.  I could make a screenshot too, however my PDF reader displays lots of whitespace between pages so I think it's probably easiest if you just look at the PDFs themselves.

[main_master.pdf](https://github.com/user-attachments/files/30463882/main_master.pdf)
[main_fix.pdf](https://github.com/user-attachments/files/30463880/main_fix.pdf)

david-christiansen:
Thanks, that really helped!

PR #924, opened 2026-07-13
fix: do not include leading file contents in blog genre `lean` blocks as whitespace

### LINKED ISSUE #923: First Lean codeblock is corrupted
The first Lean codeblock is corrupted and contains document parts from above the codeblock. 

Image

**Reproduction**
- Checkout verso-templates v4.31.0
- Open `verso-templates\blog-features\blog`
- Comment out `-- import Blog.Posts.Comparison` and `-- Blog.Posts.Comparison`  in `Posts.lean` and `Main.lean`. 
- Run `lake build`
- Run `lake exe generate-blog`
- Run `py -m http.server -d _site` and view the http://localhost:8000/blog/2025-7-19-making-fibonacci-fast-and-proving-its-correct/

whxvd:
https://github.com/leanprover/lean4/pull/12662 changed the module parser to by default include the text preceding the first token in a Lean file as whitespace belonging to that token. That parser infrastructure also is used for `lean` blocks in the blog genre. The change in the PR had the effect that the first `lean` block for an `exampleContext` always included the whole preceding Verso file contents as leading whitespace at the first command token.

Fixes #923.

### DIFF
--- a/src/verso-blog/VersoBlog.lean
+++ b/src/verso-blog/VersoBlog.lean
@@ -480,7 +480,7 @@ def lean : CodeBlockExpanderOf LeanBlockConfig
     -- Process with empty messages to avoid duplicate output
     let s ←
       withTraceNode `Elab.Verso.block.lean (fun _ => pure m!"Elaborating commands") <|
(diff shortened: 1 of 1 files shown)


david-christiansen:
Thank you!

PR #932, opened 2026-07-28
feat: produce one-sided LaTeX output by default

medovina:
Produce one-sided LaTeX output by default, which is best for online viewing, probably the most common use case.  Also added an option `--twoside` for two-sided output intended for printing.

### DIFF
--- a/src/verso-manual/VersoManual.lean
+++ b/src/verso-manual/VersoManual.lean
@@ -220,7 +220,6 @@ structure OutputConfig where
   emitHtmlMulti : EmitHtml := .immediately
   wordCount : Option System.FilePath := none
   draft : Bool := false
-  twoside : Bool := false  -- two-sided output for printing
   /-- Be verbose while generating output -/
   verbose : Bool := false
 deriving ToJson, FromJson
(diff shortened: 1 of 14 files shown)


src/verso-manual/VersoManual.lean
```diff
@@ -220,6 +220,7 @@ structure OutputConfig where
   emitHtmlMulti : EmitHtml := .immediately
   wordCount : Option System.FilePath := none
   draft : Bool := false
+  twoside : Bool := false  -- two-sided output for printing
```
david-christiansen: I think this is the right way to do it, but it should probably be on [`TeXConfig`](https://github.com/leanprover/verso/blob/637d1288c590366c177c465916177817d33fa485/src/verso-manual/VersoManual/TeX/Config.lean#L13) rather than here. This structure is more for selecting output formats to produce, rather than setting the details of any one output format. And comments like this should be docstrings instead, so they show up in hovers and the like.

david-christiansen:
Thanks!

I suggested moving the config field to a different location.

Could you please also add one golden test that ensures that the option is propagated correctly to the final TeX file?

src/verso-manual/VersoManual.lean
medovina: One possible advantage of having this field in `OutputConfig` is that it could also apply to future output formats that support one- or two-sided output, e.g. Typst.  But that's pretty speculative so sure, I'm happy to move it to `TeXConfig`.

medovina:
I've made the requested changes, so this should be ready for review again.

david-christiansen:
Thanks!

PR #936, opened 2026-07-31
Fix multi-page manual TOC links

codeguru42:
Emit explicit index.html targets for manual multi-page navigation so TOC and nav links work when opened directly from disk.

Add a regression test that builds the demo textbook and checks the generated html-multi links.


david-christiansen:
Verso output is not intended to be opened directly from disk. Various JavaScript features will not work correctly in that context, in addition to these links.

Thanks for the contribution, but I think this is out of the scope of what we want to support. Instead, please use a local web server (as described in the README).

codeguru42:
@david-christiansen thanks for your consideration. Out of curiosity what JavaScript features don't work with `file://` protocol?

PR #937, opened 2026-07-31
fix: deduplicate auto-generated heading tags in freshTag

### LINKED ISSUE #897: Automatically generated tags for headings are not deduplicated
## Summary

When automatically generated tags for headings collide, `freshTag` should append a suffix to make them unique. Instead, this deduplication does not work, resulting in a "Duplicate tag" error during manual traversal.

In particular, this issue is easy to encounter with non-ASCII headings. Every unsupported character is converted to `___`, so completely different headings can still collide if they contain the same number of unsupported characters, such as Japanese characters.

This occurs in Verso v4.29.0 and on the current `main` branch.

## Steps to reproduce

Create the following Lean file in a Verso project:
```lean
import VersoManual
open Verso.Genre Manual

def main := manualMain <| Verso.Doc.VersoDoc.toPart <| #doc (Manual) "Test" =>
# Same
First.

# Same
Second.
```

Then run `lake exe`.

## Expected behavior

Traversal succeeds. The generated tags are unique, for example `Test--Same` and `Test--Same-0`.

## Actual behavior

`lake exe` exits with status 1 and reports:
```text
Duplicate tag '_private.VersoManual.Basic.0.Verso.Genre.Manual.Tag.internal "Test--Same"'
An error was encountered!
```

`lake build` alone does not expose the problem because it compiles the document without running traversal.

## Cause

In `src/verso-manual/VersoManual/Basic.lean`, `freshTag` checks whether the `String` candidate is already present:
https://github.com/leanprover/verso/blob/f91d46ca8b63be638b2f4fef2b1011012c540a41/src/verso-manual/VersoManual/Basic.lean#L420

Because `attempt` is a `String`, the `Coe String Tag` instance converts it to `Tag.provided attempt`. However, generated tags are inserted as `Tag.internal`:
https://github.com/leanprover/verso/blob/f91d46ca8b63be638b2f4fef2b1011012c540a41/src/verso-manual/VersoManual/Basic.lean#L424

The lookup therefore cannot find an existing generated tag. This appears to be a regression introduced by commit https://github.com/leanprover/verso/commit/e842684798df452fa30c7528e49be21b7a5d682c. That commit changed `tagStr` from producing a `Tag.internal` value to producing a plain `String`. The insertion was updated to wrap the string in `Tag.internal`, but the preceding lookup was not.

## Suggested fix

Check for the same tag variant that is inserted:
```lean
if (← get).tags.contains (Tag.internal attempt) then
```

Chessing234:
## Summary
- `freshTag` looked up candidates via `tags.contains attempt`, which coerced the `String` to `Tag.provided`, while generated tags are stored as `Tag.internal`.
- Lookup now uses `Tag.internal attempt` so colliding headings get a numeric suffix instead of failing with "Duplicate tag".

Fixes #897.

## Test plan
- [ ] Manual with two `# Same` headings: `lake exe` succeeds and assigns unique tags
- [ ] Existing unique headings still get the same slug as before

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/verso-manual/VersoManual/Basic.lean
+++ b/src/verso-manual/VersoManual/Basic.lean
@@ -417,7 +417,9 @@ def freshTag [Monad m] [MonadStateOf TraverseState m] (hint : String) (id : Inte
   let mut numPart : Option Nat := none
   repeat
     let attempt := tagStr strPart numPart
-    if (← get).tags.contains  attempt then
+    -- Lookup must use `Tag.internal`: `attempt` is a `String`, and the `Coe String Tag`
+    -- instance would otherwise search for `Tag.provided`, so generated tags never collide.
+    if (← get).tags.contains (Tag.internal attempt) then
       numPart := some <| numPart.map (· + 1) |>.getD 0
     else break
   let tag := tagStr strPart numPart


david-christiansen:
It's unclear to me that there's a problem here. Can you please at least construct a test that fails with the current code and succeeds with this proposed change, as well as updating the PR description to something more informative for future readers with a bit more "forest" and less "trees"?

david-christiansen:
Closing due to lack of response. Please feel free to re-open as needed.

PR #938, opened 2026-07-31
docs: escape &amp; in HTML entity docstrings

### LINKED ISSUE #898: `"&amp;"` in docstring of `Html.text` is not rendered literally
In the docstring of `Html.text` in `src/verso/Verso/Output/Html.lean`:

```
If `escape` is `true`, then characters such as `'&'` are escaped
to entities such as `"&amp;"` during rendering.
```

At https://verso.lean-lang.org/doc/latest/Output-Formats/#Verso___Output___Html___text, the `"&amp;"` is appears as `"&"`. Expected: `"&amp;"`.

Chessing234:
## Summary
- Double-escape `&amp;` in `Html.text` / `Entities` docstrings so rendered docs show the literal entity name instead of `&`.

Fixes #898.

## Test plan
- [ ] Check rendered docs for `Html.text` show `"&amp;"` literally

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/verso/Verso/Output/Html.lean
+++ b/src/verso/Verso/Output/Html.lean
@@ -31,7 +31,7 @@ A representation of HTML, used to render Verso to the web.
 public inductive Html where
   /--
   Textual content. If `escape` is `true`, then characters such as `'&'` are escaped to entities such
-  as `"&amp;"` during rendering.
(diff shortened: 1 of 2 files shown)


david-christiansen:
This is not needed within backticks in Markdown, and not anywhere in Verso.

PR #939, opened 2026-07-31
doc: fix "paramter" and "correponding" typos

Chessing234:
This PR fixes two typos in documentation:
- "paramter" → "parameter" in the manuals user guide
- "correponding" → "corresponding" in the README


david-christiansen:
I took the liberty of updating the description, because that's what ends up as the commit message after the squash, as well as the title convention.

Thank you!

PR #946, opened 2026-08-06
feat: unorphan modules

Vtec234:
This PR ensures no library modules in Verso* are orphaned (existing under the library folder but not imported by the root module), and adds a check against that in CI. This makes `precompileModules`/`shared` targets work downstream, though we don't enable either for now - more benchmarking is needed.


david-christiansen (approved):
Deleting the old incrementality support is fine. When we re-add that, we can always go back in time if it's truly valuable.

PR #956, opened 2026-08-14
feat: add the Errata test framework

david-christiansen:
Errata is a test framework, designed to be something we can extract to its own repo after we get some experience with it here (so it can be used e.g. in verso-slides and lean-sqlite).

In Errata, tests are marked by the `@[test]` attribute. Their docstring and source range are saved for failure reporting. The value of a test can have any type with an `IsTest` instance. Some machinery in the Lakefile enumerates all tests, providing them to the runner.

Elaboration-time tests can also be implemented using `#test_msgs` and `#test_guard`, which are versions of `#guard_msgs` and `#guard` that run the compile-time test but save the result as a test case for reporting together with the rest of the tests.

Errata also supports saving JUnit XML, which various GitHub actions can conveniently display for us.


lakefile.lean
```diff
@@ -144,6 +145,224 @@ lean_exe «verso-tests» where
...
+  srcDir := errataRunnerDir
+  roots := #[`ErrataDiscovered]
+  needs := #[errataSelection]
+
+-- The generated, discovered test runner. Its source is written by the Errata test driver.
+lean_exe «errata-runner» where
+  root := `ErrataRunnerMain
+  srcDir := errataRunnerDir
+  supportInterpreter := true
+  needs := #[errataSelection]
```
robsimmons, line 184: Make this exe more internally named (errata-internal-runner) and/or create some facilitates where it notices it is being invoked unexpectedly and gives meaningful "don't do that" feedback.

src/errata/Errata/Discovery.lean
```diff
@@ -0,0 +1,144 @@
...
+when the attribute is applied; the declaration's line and column are recovered later, once the
+declaration ranges are available.
+-/
+structure TestDecl where
+  /-- The test declaration's name. -/
+  name : Name
+  /-- The source file that defines the test. -/
+  file : String
+  /-- The test's docstring, rendered as Markdown, captured when the attribute is applied. -/
+  docstring? : Option String := none
```
robsimmons: how does this work with verso.doc true?

src/errata/Errata/Here.lean
```diff
@@ -0,0 +1,32 @@
...
+@[term_elab hereStx]
+meta def elabHere : TermElab := fun _stx _expectedType? => do
+  let ref ← getRef
+  let fileMap ← getFileMap
+  let startPos := fileMap.toPosition (ref.getPos?.getD 0)
+  let endPos := fileMap.toPosition (ref.getTailPos?.getD (ref.getPos?.getD 0))
+  let file ← getFileName
+  elabTerm (← `(Errata.Location.mk $(quote file)
+      (Lean.Position.mk $(quote startPos.line) $(quote startPos.column))
+      (Lean.Position.mk $(quote endPos.line) $(quote endPos.column)))) none
```
robsimmons: does this make sense for avoiding-recusrively-elabterrming?

```suggestion
  let startPos := ref.getPos?.getD 0
  let endPos := ref.getTailPos?.getD (ref.getPos?.getD 0)
  let file ← getFileName
  return mkApp3 (mkConst ``Errata.Location.mk)
    (toExpr (← getFileName))
    (toExpr (fileMap.toPosition start))
    (toExpr (fileMap.toPosition stop))
```

src/errata/Errata/Report.lean
```diff
@@ -0,0 +1,321 @@
...
+/--
+Drops the characters XML 1.0 forbids even when escaped: those below {lit}`U+0020` other than tab,
+newline, and carriage return, and the noncharacters {lit}`U+FFFE` and {lit}`U+FFFF`.
+-/
+private def dropXmlForbidden (s : String) : String :=
+  s.foldl (init := "") fun acc c =>
+    if c == '\uFFFE' || c == '\uFFFF' then acc
+    else if c == '\t' || c == '\n' || c == '\r' || Nat.ble 0x20 c.toNat then acc.push c
+    else acc
+
```
robsimmons, line 142: Alternative: replace with `\uFFFD`.

src/errata/Errata/Result.lean
```diff
@@ -0,0 +1,176 @@
...
+
+/-- Whether each test's results are truncated after a cap at this verbosity. -/
+def Verbosity.truncates : Verbosity → Bool
+  | .quiet => true
+  | .silent | .verbose | .superVerbose => false
+
+/-- Whether every result's docstring is shown, not only those of failures and errors. -/
+def Verbosity.showsAllDocstrings : Verbosity → Bool
+  | .superVerbose => true
+  | .silent | .quiet | .verbose => false
```
robsimmons:
```suggestion
/-- Whether each test's results are truncated after a cap at this verbosity. -/
def Verbosity.truncates : Verbosity → Bool
  | .silent => false -- irrelevant
  | .quiet => true
  | .verbose | .superVerbose => false

/-- Whether every result's docstring is shown, not only those of failures and errors. -/
def Verbosity.showsAllDocstrings : Verbosity → Bool
  | .silent | .quiet | .verbose => false
  | .superVerbose => true
```

Take this or leave it, but i'd fint it nice to keep this in order, and the silent one seemed wrong at first

src/errata/Errata/Result.lean
```diff
@@ -0,0 +1,176 @@
...
+-/
+structure Location where
+  /-- The source file that contains the span. -/
+  file : String
+  /-- The start of the span. -/
+  startPos : Lean.Position
+  /-- The end of the span. -/
+  endPos : Lean.Position
+deriving Repr, Inhabited, BEq, DecidableEq
+
```
robsimmons, line 60: Consider using Lean.Lsp.Location. (Not sure it's a good idea! LSP locations are zero based! But it's the right type and already knows how to do toJson and stuff.)

src/errata/Errata/TestM.lean
```diff
@@ -0,0 +1,291 @@
...
+
+/-- Writes a file, creating all parent directories if necessary. -/
+def writeFile (path : System.FilePath) (contents : String) : IO Unit := do
+  if let some parent := path.parent then IO.FS.createDirAll parent
+  IO.FS.writeFile path contents
+
+/-- Writes a binary file, creating all parent directories if necessary. -/
+def writeBinFile (path : System.FilePath) (contents : ByteArray) : IO Unit := do
+  if let some parent := path.parent then IO.FS.createDirAll parent
+  IO.FS.writeBinFile path contents
```
robsimmons: These belong in an upstream library and not in the Errata namespace. I think the small number of places where this is used to get the createDirAll behavior should just be inlined.

src/errata/Errata/TestM.lean
```diff
@@ -0,0 +1,291 @@
...
+private def splitUtf8Tail (bytes : ByteArray) : ByteArray × ByteArray := Id.run do
+  for back in [1 : 4] do
+    if back > bytes.size then break
+    let i := bytes.size - back
+    if let some len := utf8SeqLength bytes[i]! then
+      if i + len > bytes.size then
+        return (bytes.extract 0 i, bytes.extract i bytes.size)
+      else
+        break
+  return (bytes, .empty)
```
robsimmons, line 132:
```suggestion
private def splitUtf8Tail (bytes : ByteArray) : ByteArray × ByteArray := Id.run do
  for back in [1 : 4] do
    if back > bytes.size then break
    let i := bytes.size - back
    if h : bytes[i]!.IsUTF8FirstByte then
      let len := bytes[i]!.utf8ByteSize h
      if i + len > bytes.size then
        return (bytes.extract 0 i, bytes.extract i bytes.size)
      else
        break
    else
      break
  return (bytes, .empty)
```

src/errata/Errata/TestM.lean
```diff
@@ -0,0 +1,291 @@
...
+The number of bytes in the {lit}`UTF-8` sequence a lead byte introduces, or {name}`none` for a
+continuation or invalid byte.
+-/
+private def utf8SeqLength (b : UInt8) : Option Nat :=
+  if b &&& 0x80 == 0 then some 1
+  else if b &&& 0xE0 == 0xC0 then some 2
+  else if b &&& 0xF0 == 0xE0 then some 3
+  else if b &&& 0xF8 == 0xF0 then some 4
+  else none
+
```
robsimmons:
```suggestion
```

Not needed if you use existing UInt8 functions

src/errata/Errata/Assertions.lean
```diff
@@ -0,0 +1,83 @@
...
+def assertSome {α} (value : Option α)
+    (loc : Location := by exact here%) : TestM α :=
+  match value with
+  | some v => pure v
+  | none => throw { message := "expected some, got none", location? := some loc }
+
+/-- Asserts that an option is present, without inspecting its contents. -/
+def assertIsSome {α} (value : Option α)
+    (loc : Location := by exact here%) : Test :=
+  discard (assertSome value loc)
```
robsimmons, line 86: This doesn't get used in the adaptation, and I'd be tempted to just use `let _ := assertSome v` instead if it ever came up.

src/errata/Errata/usage.txt
```diff
@@ -0,0 +1,12 @@
+Errata test runner
+
+Usage:
+  lake run Errata.run                                      run every test in the package
+  lake run Errata.run LIBRARY...                           run the tests in the given libraries
+  lake run Errata.run LIBRARY... --test-options OPTION...  pass runner options after the marker
+
+Tokens before `--test-options` name libraries. A library is a bare `Library` in this package or a
+`package/Library` reaching into a dependency. Everything after the marker goes to the test runner.
```
robsimmons: would still somewhat prefer `--` to `--test-options` but it's a rather weak preference

robsimmons (changes requested):
IsTest is very nice as an interface, I feel like I’m still learning this pattern and this is a good instance of it.

I think if we were doing a first version of an independent-release, then I’d argue against `#test_msgs` as confusing relative to the existing `#guard_msgs` which already works fine. I get there are advantages to `#test_msgs` — and I’m definitely not arguing against it in this PR, but I don't think they're worth the complexity they bring in that iteration. We'll be able to consider this in verso and downstream projects going forward, though.

Things I think should be changed, but I could be pretty easily argued down about any of them (In particular, if you want to file a "do this later" issue and go ahead and merge I'm onboard.)
 - the `.skip` mechanism in `Suppressed` (Errata.Report) and `Status` (Errata.Result) should be removed, it doesn’t seem to be exercised in #957, and skipping tests in general is a fraught subject with non-obvious design choices.
 - I only noted writeBinFile and writeFile, but there’s a couple other things in Errata.TestM that really belong not-in-errata or maybe in an Errata.Private.Util namespace.
 - The captured output doesn’t make it to JUnit (file as a ticket and do later?)
 - This fails when building ErrataDiscovered. I think you probably want it to succeed?
    ```
    @[test] unsafe def unsafeProbe : Bool := true 
    ```
 - I think Context.writeOutput is currently dead code. the simplificaiton of cutting it out, and maybe adding it back in later, might be better than wiring it up right now?
 - I don't think `-wfail` works — on #957, `DemoSite.Blog.Conditionals` has warnings but doesn't fail if you run with `-wfail`. Remove and consider adding back in later, perhaps?

lakefile.lean
david-christiansen: Done! Thanks!

src/errata/Errata/Discovery.lean
david-christiansen: It has Lean render it to Markdown when it captures the docstring, just as is done for hovers.

I added a test that makes sure this works right.

src/errata/Errata/Report.lean
david-christiansen: Good call. Fixed.

src/errata/Errata/Result.lean
david-christiansen: I did a slightly more radical version. Thanks!

src/errata/Errata/TestM.lean
david-christiansen: I agree with the diagnosis but not the treatment.

Instead, I made them private. Having these helpers is a big footgun safety switch for me.

src/errata/Errata/TestM.lean
david-christiansen: Good point! Deleted.

src/errata/Errata/TestM.lean
david-christiansen: This suggestion didn't work for three-byte sequences, but using `IsUTF8FirstByte` is a nice improvement. Fixed version pushed.

src/errata/Errata/usage.txt
david-christiansen: I still don't want to do this, based on the lived experience of typing `--` instead of `-- --` repeatedly and being temporarily confused. Otherwise I'd totally be on board.

david-christiansen:
Thanks for the detailed review!

Here's point-by-point:
 
> I think if we were doing a first version of an independent-release, then I’d argue against `#test_msgs` as confusing relative to the existing `#guard_msgs` which already works fine. I get there are advantages to `#test_msgs` — and I’m definitely not arguing against it in this PR, but I don't think they're worth the complexity they bring in that iteration. We'll be able to consider this in verso and downstream projects going forward, though.

I still think this is important to keep. The point of it is that it creates a run-time witness for an elaboration-time test's result, which allows the test to be included in a report of test results at run time. I can see "this #eval failed, and so did this @[test] assertion" without having to first resolve all of the first category so the build will succeed before I can resolve any from the second category. Sometimes I want to deal with test failures by feature or code region rather than by compiler phase. But time will tell.

> Things I think should be changed, but I could be pretty easily argued down about any of them (In particular, if you want to file a "do this later" issue and go ahead and merge I'm onboard.)
> 
> * the `.skip` mechanism in `Suppressed` (Errata.Report) and `Status` (Errata.Result) should be removed, it doesn’t seem to be exercised in #957, and skipping tests in general is a fraught subject with non-obvious design choices.

Indeed! It's gone now.

> * I only noted writeBinFile and writeFile, but there’s a couple other things in Errata.TestM that really belong not-in-errata or maybe in an Errata.Private.Util namespace.

What else is there to remove? I'm not seeing any other obviously-not-germane-to-testing public functions.

> * The captured output doesn’t make it to JUnit (file as a ticket and do later?)

I just fixed this one. Thanks!

> * This fails when building ErrataDiscovered. I think you probably want it to succeed?
>       ```
>       @[test] unsafe def unsafeProbe : Bool := true 
>       ```

Yep. Fixed!

> * I think Context.writeOutput is currently dead code. the simplificaiton of cutting it out, and maybe adding it back in later, might be better than wiring it up right now?

This is used in #959.

> * I don't think `-wfail` works — on #957, `DemoSite.Blog.Conditionals` has warnings but doesn't fail if you run with `-wfail`. Remove and consider adding back in later, perhaps?

Those are Lake's build warnings, not warnings that occur while executing the tests. This feature is used for test framework warnings like unused command-line parameters or tests that aren't included in the globs, which I think are important to notice. I edited the potentially-confusing docstring and added a test to make sure it works.

robsimmons (approved):
I think this is definitely on-balance worth landing with these changes, and any follow-ups can be follow-up work. Thanks!

david-christiansen:
Thanks @robsimmons!

PR #957, opened 2026-08-14
chore: migrate the test suite to Errata

david-christiansen:
This rearchitects our test suite into a proper testing framework. Before, our tests were all essentially ad hoc. Each kind of test would be an IO action that might throw, and there were some rough conventions, but output reporting was inconsistent and sometimes hard to understand. Tests were noisy as well, which made it hard to find reasons for failure at a glance.

Now the suite runs on Errata: `lake test` discovers every `@[test]` in the package, reports results uniformly, and CI displays the JUnit XML report.


.github/workflows/test-imports.yml
```diff
@@ -1,3 +1,8 @@
+# This check is retained only to satisfy the required "Check all test modules are imported" status
+# check on the protected branch. It is a no-op on this layout: test modules now live under
+# src/tests/VersoTests and are discovered by globbing, and src/tests/Tests no longer exists, so the
+# scan below finds nothing. Coverage is instead enforced by the test driver's unreachable-modules
+# warning during `lake test`. Remove this workflow (and drop the required check) once merged.
```
robsimmons, line 5: (This is fine but) you could add the removal to the PR stack, I guess?

robsimmons (approved):
One of the spot-checks I did was on the new LzCompress module. I'd note that, near as I can tell, this test has no callers and isn't run. So, good case in point for Errata!

.github/workflows/test-imports.yml
david-christiansen: If we do, then it fails CI. So I think this is just a TODO for whoever merges: first merge, then delete the required check, then do a quick followup to delete this file.

PR #959, opened 2026-08-14
feat: run Errata tests interactively in a widget

david-christiansen:
Tests can additionally be run interactively using a widget. When the text cursor is on a test's source span, the InfoView offers a "run" button that runs the test in a separate process, streaming its output as it is produced and killing the process if the run is cancelled.

### DIFF
--- a/src/errata-tests/ErrataTests.lean
+++ b/src/errata-tests/ErrataTests.lean
@@ -155,6 +155,62 @@ def reportMarkdown : Test := do
   assertContains "expected 1\nactual 2" md
   assertContains "Summary by module" md
 
+/-- `runValue` reports a passing value as passed. -/
+@[test]
+def runOnePasses : Test := do
+  let o ← runValue default (pure () : Test)
+  assertEq "passed" o.status
+
+/-- `runValue` reports a failing value as failed and carries its message. -/
+@[test]
+def runOneFails : Test := do
+  let o ← runValue default (TestResult.fail { message := "boom" })
+  assertEq "failed" o.status
+  assertEq (some "boom") o.message?
+
+/-- `runValue` reports a skipped value as skipped. -/
+@[test]
+def runOneSkips : Test := do
+  let o ← runValue default (TestResult.skip "later")
+  assertEq "skipped" o.status
+
+/-- A failing run surfaces its captured output in the outcome. -/
+@[test]
+def runOneCapturesOutput : Test := do
+  let o ← runValue default (do IO.println "trace line"; failHere "nope" : Test)
+  assertEq "failed" o.status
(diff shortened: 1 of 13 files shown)


src/errata/Errata/Discovery.lean
```diff
@@ -58,6 +60,70 @@ meta def recordTest (decl : Name) : AttrM Unit := do
...
+meta def widgetRangeSyntax (decl : Name) (attrStx : Syntax) : AttrM Syntax := do
+  let fileMap ← getFileMap
+  if let some ranges ← findDeclarationRanges? decl then
+    let stx ← rangeSyntax (fileMap.ofPosition ranges.range.pos) (fileMap.ofPosition ranges.range.endPos)
+    return stx
+  let some attrPos := attrStx.getPos? | return attrStx
+  let lineStart := fileMap.ofPosition ⟨(fileMap.toPosition attrPos).line, 0⟩
+  let inputCtx := Parser.mkInputContext fileMap.source (← getFileName)
+  let pmctx : Parser.ParserModuleContext := { env := ← getEnv, options := ← getOptions }
+  let (cmdStx, _, _) := Parser.parseCommand inputCtx pmctx { pos := lineStart } {}
```
Vtec234, line 116: Re-parsing is a bit of waste - `AttrM` really should have easier access to the syntax of the marked declaration; but that's a core change AFAICT. There's an attribute with the same issue in verso-blueprint.

src/errata-tests/ErrataTests.lean
```diff
@@ -15,6 +15,7 @@ import all Errata.FS
 import all ErrataTests.Fixture
 import all ErrataTests.Fixture.Sub
 import all ErrataTests.Docstrings
+import all ErrataTests.WidgetInteractive
```
Vtec234: This doesn't exist.

Vtec234, line 18: This doesn't exist.

david-christiansen: I forgot to `git add` it - sorry!

That's basically a file for manually testing the widget behavior that isn't automatically testable.

Vtec234 (approved):
One confusing behavior I found: nested stdout/err boxes do not have a copy button, whereas the copy button on the top box actually copies all the output including nested outputs. Since the boxes are visually separated, this feels surprising.

https://github.com/user-attachments/assets/0cea4cb6-00f6-4f7e-bb58-875f3ec78db7

Vtec234:
I am also wondering, should `assertTrue condition message` display `message` in the widget on failure? It currently doesn't.

david-christiansen:
Copy button and messages fixed.

Thanks again!

PR #961, opened 2026-08-15
fix: use saved environment for verso-blog's inline Lean

berberman:
This PR fixes the bug that verso-blog's inline Lean role didn't use saved example environment to elaborate terms by wrapping the elaboration in `try setEnv env ... finally setEnv oldEnv`.

No-Changelog: this is a simple bugfix


david-christiansen:
Thanks!

PR #966, opened 2026-08-21
chore: bump toolchain to v4.34.0-rc2

kim-em:
No-Changelog: routine toolchain bump for the v4.34.0-rc2 release.

PR #967, opened 2026-08-21
fix: sync lean-toolchain for test projects that build against Verso

kim-em:
This PR keeps `test-projects/literate-config` and `test-projects/literate-multi-root` on the root toolchain when it is bumped, so the SubVerso consistency check no longer fails on every toolchain bump.

`update-subverso.sh` iterated over `lake-manifest.json` files, so it never reached those two projects: they require Verso by path and therefore have no manifest. The projects that must track the root toolchain are now listed in `test-projects/root-toolchain-projects.txt`, read by the updater, the consistency check, and the `!consistent` fixer alike, so the three cannot drift apart. Every other test project depends on SubVerso from Git and stays pinned to its own toolchain.

The fixer resolves the list before checking out the pull request, since it runs with write access and should not take instructions from pull request content.

No-Changelog: release tooling, with no effect on Verso's users.

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/consistent-pr-comment.yml
+++ b/.github/workflows/consistent-pr-comment.yml
@@ -63,6 +63,28 @@ jobs:
               with:
                   token: ${{ secrets.GITHUB_TOKEN }}
 
+            - name: Resolve toolchain sync targets
+              # Read the list from the default-branch checkout, before the pull
+              # request's own code is on disk. This job can write to the
+              # repository, so it must not take instructions from the PR.
+              run: |
+                  list=test-projects/root-toolchain-projects.txt
+                  mapfile -t names < <(grep -vE '^[[:space:]]*(#|$)' "$list")
+                  if [ "${#names[@]}" -eq 0 ]; then
+                    echo "Error: no projects listed in $list" >&2
+                    exit 1
+                  fi
+                  for name in "${names[@]}"; do
+                    case "$name" in
+                      */* | . | ..)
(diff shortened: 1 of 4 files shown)


kim-em:
:bot: says:

Handing this over rather than merging it during the release. @david-christiansen, your call entirely.

Background: the v4.34.0-rc2 toolchain bump (#966) failed the SubVerso consistency check because `test-projects/literate-config` and `test-projects/literate-multi-root` were still on rc1. `update-subverso.sh` iterates over `lake-manifest.json` files, and those two require Verso by path so they have no manifest, meaning the updater never reached them.

Two things worth your eye rather than a rubber stamp:

* The projects that must track the root toolchain now live in `test-projects/root-toolchain-projects.txt`, read by the updater, the consistency check, and the `!consistent` fixer alike. That is a policy decision as much as a fix, and you may prefer a different mechanism.
* `consistent-pr-comment.yml` is an `issue_comment` workflow with `contents: write`. It now resolves that list from the default-branch checkout *before* `gh pr checkout`, so the privileged job never reads or executes anything the pull request controls, and it validates each entry is a bare directory name and refuses symlinks. An earlier draft of mine had it running a script from the PR branch, which would have been a privilege-escalation hole; that is gone, but the workflow is worth a second pair of eyes regardless.

Verified by reproducing the actual rc2 failure and watching the updater repair it, with the four test projects deliberately pinned to v4.31.0 left untouched, and by checking the empty-list, missing-list and path-traversal cases all fail loudly rather than passing vacuously.

PR #969, opened 2026-08-21
fix: gracefully handle unknown -D options in verso-literate

### LINKED ISSUE #968: leanOptions handling potentially brokeen
Adding 
```
[leanOptions]
weak.linter.unusedVariables = false

```
to the `lakefile.toml`
causes build failures in the literate faces, see commits in an example repo I made:

https://github.com/mo271/WeakLiterate

I have a potential fix prepared, will make a pr shortly.

mo271:
When `lake lint` runs, it passes linter overrides to downstream executables by prefixing the options with `weak.` (e.g., `-Dweak.linter.style.openClassical=true`).

Previously, `parseDOption` strictly relied on `getOptionDecl name` to find the expected type for an option. If the option was unknown (e.g. because of the `weak.` prefix or because it hasn't been registered yet), `getOptionDecl` threw an internal exception, causing `verso-literate` to crash.

This patch mirrors the behavior of `setConfigOption` in core Lean (`src/Lean/Shell.lean`). By using `(← getOptionDecls).find? name`, we can check if the option exists. If it does, we parse it according to its type. If it doesn't, we gracefully fall back to storing it as a string in the `Options` map and defer validation to the elaborator, preventing the crash.

fixes #968, see that issue for an example repo

### DIFF
--- a/src/verso-literate/VersoLiterateMain.lean
+++ b/src/verso-literate/VersoLiterateMain.lean
@@ -709,31 +709,33 @@ private def parseDOption (arg : String) (opts : Options) : IO Options := do
   | [name, value] =>
     let name := String.toName name.copy
     let value := value.copy
-    let decl ← getOptionDecl name
-    match decl.defValue with
-    | .ofBool _ =>
-      match value with
-      | "true" => return opts.setBool name true
-      | "false" => return opts.setBool name false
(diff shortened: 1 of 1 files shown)


mo271:
the diff might appear larger than it really is due to whitespace, looks less scary with `git diff -w`

PR #970, opened 2026-08-21
perf: batch-highlighting of code

david-christiansen:
This PR migrates more of Verso to the highlightMany API of SubVerso, allowing it to share more caches when highlighting longer chains of commands.

There is also a change to a proof. With this change, the library works on Lean 4.33 as well, which is important for a downstream user who needs these changes but can't get to latest Lean release candidate.

No-Changelog: this is a dep bump


src/verso-manual/VersoManual/InlineLean.lean
```diff
@@ -319,12 +327,13 @@ meta def elabCommands (config : LeanBlockConfig) (str : StrLit)
...
 
-      let mut hls := Highlighted.empty
       let nonSilentMsgs := cmdState.messages.toArray.filter (!·.isSilent)
-      let mut lastPos : String.Pos.Raw := startPos
-      for cmd in cmds do
-        hls := hls ++ (← highlightIncludingUnparsed cmd nonSilentMsgs cmdState.infoState.trees (startPos? := lastPos))
-        lastPos := (cmd.getTrailingTailPos?).getD lastPos
+      -- One batched highlighting pass shares the alias-table and info-table setup across the
+      -- whole block's commands. Unparsed regions (e.g. text skipped by parse-error recovery)
+      -- are included in the output verbatim.
```
```suggestion
      -- One batched highlighting pass shares caches between the whole block's commands.
```

PR #974, opened 2026-08-27
chore: use upstream HTML type

Vtec234:
This PR switches Verso to use the upcoming core Lean HTML type rather than defining one.

JSX-like syntax and utilites are not changed here.


doc/UsersGuide/Output/HTML.lean
```diff
@@ -26,23 +27,21 @@ tag := "output-html"
...
+{docstring Html.ofArray}
 
-{docstring Html.fromList}
+{docstring Html.ofList}
 
 {docstring Html.append}
 
 {docstring Html.visitM}
 
-{docstring Html.format}
```
Vtec234, line 44: This function has actually been deleted, since nothing in Verso uses it. Let me know if you'd rather keep it.

david-christiansen: That's fine. It used to be used in tests to get more readable output IIRC, but we can get rid of it. It wasn't correct WRT whitespace-sensitive tags anyway.

david-christiansen:
Given that lots of external codebases use this library, I think we should endeavor to provide deprecated aliases for all the names that point at the Lean versions, to make adaptation less painful.

src/verso/Verso/Output/Html.lean
```diff
@@ -21,117 +22,25 @@ public meta import Verso.Output.Html.Comments
...
+namespace Lean.Html
 
 /--
 If the HTML consists of a single tag, then the given attribute is set to the provided value. If the
 attribute already exists, then its value is replaced.
 
 If the HTML is not a single tag, no changes are made.
 -/
-public def Html.setAttribute (attr : String) (value : String) (html : Html) : Html :=
+public def setAttribute (attr : String) (value : String) (html : Html) : Html :=
```
david-christiansen, line 35: This is also used downstream, so a deprecated alias in the old namespace would be nice.

src/verso/Verso/Output/Html.lean
```diff
@@ -202,31 +70,58 @@ public def labeledGroup
...
+where
+  newline i := "\n" ++ String.ofList (List.replicate i ' ')
+  breakline tag := if breakLines && tag ∈ newlineAfter then newline indent else ""
+  breakline' tag := if breakLines && tag ∈ newlineAfter then newline (indent + 2) else ""
+  attrsAsString xs := String.join <| xs.toList.map (fun ⟨k, v⟩ => s!" {k}=\"{escapeAttr v}\"")
+  escapeAttr str := str |>.replace "&" "&amp;" |>.replace "\"" "&quot;"
 
-public section
+/-- The default `DOCTYPE` for HTML5. -/
+public abbrev doctype := "<!DOCTYPE html>"
```
david-christiansen: Does this really belong in `Lean.Html`?

src/verso/Verso/Output/Html.lean
```diff
@@ -202,31 +70,58 @@ public def labeledGroup
...
+
+end Lean.Html
+
+/-! ## JSX-like syntax -/
```
david-christiansen: Historically, this was at least as inspired by Handlebars and Django templates :-)

Vtec234:
> provide deprecated aliases for all the names that point at the Lean versions

I did this at first, but this mostly results in name resolution failures downstream. For example with `@[deprecated] abbrev Verso.Output.Html := Lean.Html`, downstream files that `open Verso Output` and `open Lean` - which, at least in Verso, most do - end up with 'ambiguous symbol' errors, or the wrong `Html` type inferred, in ways that make it harder to adapt than when we just break the API, ~~like this PR does. I am open to tricks that would work around this, if you know of any.~~

UPDATE: I think I have found the optimal thing to do using a combination of `export` and `public/protected abbrev`; see the bottom of `Verso.Output.Html`.

src/verso/Verso/Output/Html.lean
Vtec234: Moved back to `Verso.Output.Html`.

src/tests/Tests/VersoManual/Html/SoftHyphenate.lean
```diff
@@ -13,24 +13,24 @@ open Verso.Genre.Manual
 /-- info: "blahNotCode<code><a>foo&shy;Bar&shy;Baz</a></code>" -/
 #guard_msgs in
 open Verso.Output Html in
-#eval softHyphenateIdentifiers {{"blahNotCode"<code><a>"fooBarBaz"</a></code>}} |>.asString
+#eval softHyphenateIdentifiers {{"blahNotCode"<code><a>"fooBarBaz"</a></code>}} |>.render
 
-/-- info: "<code>abc.<wbr>def.<wbr>ghi.<wbr>jkl</code>" -/
+/-- info: "<code>abc.<wbr/>def.<wbr/>ghi.<wbr/>jkl</code>" -/
```
Vtec234, line 18: This diff is because we now self-terminate void tags in the upstream renderer; but the output is not valid XML in general (e.g. uses attributes without values like `<input disabled/>` which are forbidden in XML), so maybe this is a bad choice and we should go all-in on HTML syntax like the original Verso code. What do you think?

doc/UsersGuide/Output/HTML.lean
```diff
@@ -59,19 +58,12 @@ def mkList (xs : List Html) : Html :=
...
 ```leanOutput htmllist
-<ul>
-  <li>
-    A</li>
-  <li>
-    <emph>B</emph></li>
-  <li>
-    C</li>
-  </ul>
+<ul><li>A</li><li><emph>B</emph></li><li>C</li></ul>
```
Vtec234, line 66: `asString` used to say "Converts HTML into a string that's suitable for sending to browsers, but is also readable.". IMO send-to-browsers-output and readable-output have different requirements and should really be two functions; so far I only added the send-to-browsers function (`render`) upstream, but a side effect of switching to that is uglier output. What would you like to do here?

src/tests/Tests/Html.lean
```diff
@@ -93,33 +93,3 @@ Hint: Remove contents
 -/
 #guard_msgs in
   #eval show Html from {{ <br>"foo" "foo"</br> }}
-
-/--
```
Vtec234, line 97: Rendering and escaping is tested upstream now.

src/tests/Tests/VersoManual/Html/SoftHyphenate.lean
Vtec234: The community `lean-html` package [supports both HTML5 and XHTML](https://github.com/paulbutcher/lean-html/commit/9721f7888a8177f98096a570653e0e8ff0b0f9b0), while I'd explicitly prefer not to - we should be opinionated here.

PR #987, opened 2026-09-17
feat: adopt upstream HTML syntax

Vtec234:
This PR deprecates the HTML syntax `{{ .. }`, moving all its uses to new parsers proposed in Lean core.

The adaptations are largely mechanical. More meaningful changes - removed `Html` test, changed `HTML` manual chapter - are separated into standalone commits.


doc/UsersGuide/Output/HTML.lean
```diff
@@ -44,20 +43,15 @@ In Verso, they are typically produced using an embedded DSL that is available wh
...
 {docstring Html.render}
 
-HTML documents are written in double curly braces, in a syntax very much like HTML itself.
-The differences are:
- * Double curly braces escape back to Lean. This can be done for HTML elements, attribute values, or whole sets of attributes.
- * Text content is written as Lean string literals to facilitate precise control over whitespace.
- * Interpolated Lean strings (with `s!`) may be used in any context that expects a string.
 
+HTML documents are typically produced using an embedded DSL,
+accessible via the `html%{…}` literal syntax.
```
david-christiansen: Please keep the one sentence per line convention instead of paragraph-wrapping.

david-christiansen: It would also be good to describe it a bit more here (or, better, in the language reference with a link from here), esp. the locations where interpolations are allowed.

david-christiansen: Also the `...` interpolation syntax

src/verso-manual/VersoManual/Docstring/Progress.lean
```diff
@@ -145,53 +145,51 @@ public def progress.descr : BlockDescr where
...
             <pre>
-              {{ undocTactics.map ("{docstring " ++ ·.toString ++ "}\n") |>.toList |> String.join }}
+              { undocTactics.map ("{docstring " ++ ·.toString ++ "}\n") |>.toList |> String.join }
             </pre>
             <pre>
-              "```exceptions\n"
-              {{ undocTactics.map (·.toString ++ "\n") |>.toList |> String.join }}
-              "```\n"
+              ```exceptions{.text "\n"}{ undocTactics.map (·.toString ++ "\n") |>.toList |> String.join
+              }```{.text "\n"}
```
david-christiansen: With the new syntax, it's much more difficult for me to read this code and understand that it's doing the right thing. Please make sure it's tested, or at least test it yourself before a merge - I think it's probably OK but it's hard to be sure about the whitespace.

src/verso-manual/VersoManual/Bibliography.lean
```diff
@@ -139,7 +139,7 @@ private def andList (xs : Array Html) : Html :=
   else if h : xs.size = 2 then xs[0] ++ " and " ++ xs[1]
   else
     open Html in
-    (xs.extract 0 (xs.size - 1)).foldr (init := {{" and " {{xs.back}} }}) (· ++ ", " ++ ·)
+    (xs.extract 0 (xs.size - 1)).foldr (init := html%{{.text " and "}{xs.back} }) (· ++ ", " ++ ·)
```
david-christiansen: Is the HTML literal really helping us out here enough?

I would think that `Html.text " and " ++ xs.back` might be easier now that we have to worry about whitespace here.

src/verso-manual/VersoManual/License.lean
```diff
@@ -40,17 +40,16 @@ private def paragraphed (text : String) : Array String := Id.run do
...
-      {{link.map (fun url => {{<a class="link" href={{url}}>{{url}}</a>}}) |>.getD .empty}}
-      {{howUsed.map paragraphedHtml |>.getD .empty}}
-      <code class="spdx">{{identifier}}</code>
-      {{text.map textHtml}}
-    </section>}}
+  html%{<section class="license-info">
+      {.element s!"h{headerLevel}" #[] dependency
+      }{link.map (fun url => html%{<a class="link" href={url}>{url}</a>}) |>.getD .empty
+      }{howUsed.map paragraphedHtml |>.getD .empty
+      }<code class="spdx">{identifier}</code>{text.map textHtml}
```
david-christiansen: This section reminds me of % hacks used in LaTeX to avoid whitespace.

What about this instead? Seems less likely to break due to someone being confused about the formatting and "fixing" it.

```suggestion
      {.element s!"h{headerLevel}" #[] dependency ++
        (link.map (fun url => html%{<a class="link" href={url}>{url}</a>}) |>.getD .empty) ++
        (howUsed.map paragraphedHtml |>.getD .empty) ++
        html%{<code class="spdx">{identifier}</code>{text.map textHtml} }
```

src/verso-manual/VersoManual/Literate.lean
```diff
@@ -46,12 +46,12 @@ block_extension Block.literateDocstringPart (level : Nat) where
         reportError s!"Expected a block at the beginning of a docstring section"
         pure .empty
     let contents := contents.extract 1
-    pure {{
+    pure html%{
       <section>
-        {{.element s!"h{level + 1}" #[] title}}
-        {{← contents.mapM goB}}
+        {.element s!"h{level + 1}" #[] title
+        }{← contents.mapM goB}
```
david-christiansen:
```suggestion
        {.element s!"h{level + 1}" #[] title
          ++ (← contents.mapM goB)}
```

src/verso-manual/VersoManual.lean
```diff
@@ -767,22 +767,22 @@ where
...
     let pageContent := open Verso.Output.Html in
-      {{<section>
-          {{Html.titlePage titleHtml authors authorshipNote introHtml}}
-          {{bookTocHtml}}
-          {{contents}}
-        </section>}}
+      html%{<section>
+          {Html.titlePage titleHtml authors authorshipNote introHtml
+          }{bookTocHtml
+          }{contents}
```
david-christiansen:
```suggestion
          {Html.titlePage titleHtml authors authorshipNote introHtml
            ++ bookTocHtml
            ++ contents}
```
if this is to avoid whitespace. But that trick shouldn't be needed here, so why not

```suggestion
          {Html.titlePage titleHtml authors authorshipNote introHtml}
          {bookTocHtml}
          {contents}
```

src/verso-tutorial/VersoTutorial.lean
```diff
@@ -432,25 +431,25 @@ nav.local-toc .code-link.download::before {
...
+          { if live?.isSome || code?.isSome then html%{
               <div class="code-links">
-                {{if let some live := live? then
-                  {{ <a href={{live.url}} class="live code-link">"Live"</a> }}
-                  else .empty}}
-                {{if let some (url, _file) := code? then {{ <a href={{url}} class="download code-link"><code>".zip"</code></a> }} else .empty}}
+                {if let some live := live? then
+                  html%{ <a href={live.url} class="live code-link">Live</a> }
+                  else .empty
+                }{if let some (url, _file) := code? then html%{ <a href={url} class="download code-link"><code>.zip</code></a> } else .empty}
```
david-christiansen: I don't think the whitespace breaks this. If it does, maybe do a ++ instead?
```suggestion
                  else .empty}
                {if let some (url, _file) := code? then html%{ <a href={url} class="download code-link"><code>.zip</code></a> } else .empty}
```

src/verso/Verso/Doc/Html.lean
```diff
@@ -145,25 +145,25 @@ partial def Block.toHtml [Monad m] [GenreHtml g m] [TraverseBlock g] (b : Block
...
+          pure html%{
+            <dt>{← t.mapM Inline.toHtml }</dt>
+            <dd>{← d.mapM Block.toHtml }</dd>
+          }
+        }
       </dl>
-    }}
-  | .code content => return {{ <pre> {{ content }} </pre>}}
+    }
+  | .code content => return html%{ <pre>{ .text content }</pre>}
```
david-christiansen: Why do we need the explicit `.text`? Shouldn't interpolating a string do that by default?

david-christiansen:
I'd like to see the docs come in with a bit more details (the available interpolation formats, what they mean, whitespace handling). No need for a massive tome, but a few sentences on each would be very useful.

There's also a pattern I commented on a few times, where a line-leading `}{` seems to be used to avoid having a single space text node inserted in the tree. I think this is fragile and it'd be better to just have one interpolation that includes an append in these cases, because my first instinct when I see that is to move the `}` up to the preceding line.

Other than that, looks great :)

david-christiansen:
(I really wish "Close with comment" had an "are you sure" prompt - sorry about that, it was a mistake)

doc/UsersGuide/Output/HTML.lean
Vtec234: I want to add a refman chapter on JSON, HTML, the associated syntaxes, and widgets. Are you happy to defer linking back to that from Verso to another PR?

src/verso-manual/VersoManual/Docstring/Progress.lean
Vtec234: We designed the syntax to be whitespace-sensitive within content, but this is indeed footgunny for interpolations. Do you think we should rethink the rules, for instance removing whitespace between adjacent interpolations?

src/verso-manual/VersoManual/Bibliography.lean
Vtec234: It's now `html%{ and {xs.back}}`.

src/verso-manual/VersoManual/Docstring/Progress.lean
Vtec234: Ok, I have rolled out a syntax with simpler rules and force-pushed here. This section now looks like
```lean
                <pre>
                  {"```exceptions\n"}
                  { notDocumented.map (·.toString ++ "\n") |> String.join }
                  {"```\n"}
                </pre>
```
and the newline rules guarantee that no whitespace beyond what's explicitly here is added.

Vtec234:
Changed whitespace rules to make interpolation look more natural. Checked that the test projects produce identical HTML as before, and spot-checked some trickier literals for identical output also.

doc/UsersGuide/Output/HTML.lean
david-christiansen: Yes, that's fine. I just don't want our already-insufficient docs to degrade, but a temporary downgrade is ok.

